PR #1, opened 2014-09-04
Update edDraftURI, drop prevED.

anssiko:


yell0wd0g:
I think these two field links have been corrected, so I'll close this pull request.

Issue #11, opened 2015-07-13
BlobEvent constructor arguments are missing


foolip:
http://w3c.github.io/mediacapture-record/MediaRecorder.html#blob-event

As a consequence, the BlobEventInit dictionary is unused in the spec.

yell0wd0g:
Closing this issue per https://github.com/w3c/mediacapture-record/pull/25.

yell0wd0g:
Also this guy https://github.com/w3c/mediacapture-record/pull/37.

PR #25, opened 2015-12-04
Correct BlobEvent constructor

yell0wd0g:
BlobEvent does not mention the |type| or BlobEventInit constructor parameters. This pull adds those and their descriptions.

Addresses issue https://github.com/w3c/mediacapture-record/issues/11.
Implemented by Gecko [1] and Chromium [2] in this way, to the best of my knowledge.

[1] https://github.com/mozilla/gecko-dev/blob/master/dom/webidl/BlobEvent.webidl
[2] https://code.google.com/p/chromium/codesearch#chromium/src/third_party/WebKit/Source/modules/mediarecorder/BlobEvent.idl&q=blobevent%20idl&sq=package:chromium&type=cs

### DIFF
--- a/MediaRecorder.html
+++ b/MediaRecorder.html
@@ -273,15 +273,14 @@ behavior specified in this document. </p>
 
 <section id="blob-event">
     <h2>Blob Event</h2>
-    <dl title='[Constructor(DOMString type, optional BlobEventInit eventInit)] interface BlobEvent : Event' class='idl'>
-        <dd>                   
-            <dl class='parameters'>
+    <dl title='interface BlobEvent : Event' class='idl'>
+        <dt>[Constructor(DOMString type, optional BlobEventInit eventInit)]</dt>
+            <dd><dl class='parameters'>
                 <dt>DomString type</dt>
                 <dd>Associated <a href=https://dom.spec.whatwg.org/#interface-event>Event type.</a></dd>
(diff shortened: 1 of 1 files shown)


MediaRecorder.html
```diff
@@ -272,23 +272,25 @@
...
-        <h3>BlobEventInit</h3>
-  <dl title='dictionary BlobEventInit' class='idl'>
-          <dt>Blob data</dt>
-          <dd>
-            A Blob object containing the data to deliver via this event.
-          </dd>
-        </dl>
+    <h2>Blob Event</h2>
+    <dl title='interface BlobEvent : Event' class='idl'>
+        <dt>Constructor(DOMString type, optional BlobEventInit eventInit)</dt>
```
foolip, line 277: Because data attribute on Blob is not nullable, there cannot be a way to create a BlobEvent without providing a Blob. In other words, this eventInitDict (this is what it's usually called) cannot be optional, and the data member of BlobEventInit must be required.

MediaRecorder.html
```diff
@@ -272,23 +272,25 @@
...
+            <dd><dl class='parameters'>
+                <dt>DomString type</dt>
```
foolip, line 279: s/DomString/DOMString/

yell0wd0g:
sorry, my bad. Addressed those two comments in https://github.com/w3c/mediacapture-record/pull/37.

Issue #14, opened 2015-08-31
Remove RecordingErrorNameEnum and use DomException instead


yell0wd0g:
The said RecordingErrorNameEnum [1] should not exist stand alone,
since it partially overlaps DomException [2], concretely 
INVALID_STATE_ERR and SECURITY_ERR. Reuse DomException.

[1] http://www.w3.org/TR/mediastream-recording/#MediaRecorderErrorEvent
[2] http://www.w3.org/wiki/DOM/domcore/DOMException

beverloo:
This also goes for RecordingExceptionEnum, which defines "UnsupportedOption". A NotSupportedError DOMException should be used instead.

yell0wd0g:
Filed https://github.com/w3c/mediacapture-record/issues/18 for RecordingExceptionEnum::UnsupportedOption.

yell0wd0g:
Actually please ignore [1] and use [2] instead (the former is gone).

[1] http://www.w3.org/wiki/DOM/domcore/DOMException
[2] http://www.w3.org/TR/2015/WD-dom-20150618/#exception-domexception

foolip:
I would probably suggest firing a simple error event, and adding a `readonly attribute DOMException error` to the MediaRecorder interface instead. This is how [`HTMLMediaElement`](https://html.spec.whatwg.org/multipage/embedded-content.html#media-elements) and [`IDBRequest`](http://w3c.github.io/IndexedDB/#request-api) etc. do similar things.

yell0wd0g:
Can you have a look at https://github.com/w3c/mediacapture-record/pull/29 and let me know if this is closer/farther to a satisfactory conclusion? Happy to iterate.

yell0wd0g:
Ok, let's forget about https://github.com/w3c/mediacapture-record/pull/29 in favour of first tackling the DOMExceptions and then the Error Events.

Let's concentrate the Error Event conversation in bug https://github.com/w3c/mediacapture-record/issues/31, if you guys don't mind.

PR #29, opened 2015-12-05
Removing MediaRecorderErrorEvent, in favour of just a DOMException

yell0wd0g:
This is my first attempt at removing the MediaRecorderErrorEvent (https://github.com/w3c/mediacapture-record/issues/14). I included the list of potential DOMExceptions generated. 

I'm absolutely 0% sure I'm writing the right w3c lingo, please review.

### DIFF
--- a/MediaRecorder.html
+++ b/MediaRecorder.html
@@ -13,263 +13,162 @@
   </head>
 
 
-  <body>
-        <section id="abstract">
-                <p>This document defines a recording API for use with MediaStreams as defined in
-                        [[!GETUSERMEDIA]]</p></section>
+<body>
+ <section id="abstract">
+  <p>This document defines a recording API for use with MediaStreams as defined in [[!GETUSERMEDIA]]</p>
+ </section>
 
-  <section id="sotd">
+ <section id="sotd">
+  <p>This document is not complete. It is subject to major changes and, while early experimentations are encouraged, it is therefore not intended for implementation. The Media Capture Task Force expects this specification to evolve significantly based on: </p>
+  <ul>
+   <li>Privacy issues that arise when capturing media.</li>
+   <li>Technical discussions within the task force.</li>
+   <li>Experience gained through early experimentations.</li>
+   <li>Feedback received from other groups and individuals.</li>
+  </ul>
+ </section>
 
-    <p>This document is not complete. It is subject to major changes and, while
-    early experimentations are encouraged, it is therefore not intended for
-    implementation.
-    The Media Capture Task Force expects this specification to evolve
-    significantly based on: </p>
-
-    <ul>
-      <li>Privacy issues that arise when capturing media.</li>
-
-      <li>Technical discussions within the task force.</li>
-
-      <li>Experience gained through early experimentations.</li>
-
-      <li>Feedback received from other groups and individuals.</li>
-    </ul>
-
-
-</section>
-
-
-<section id="overview" class="informative"><h2>Overview</h2>
-
-  <p>This API attempts to make basic recording very simple, while still allowing for more complex use cases.  In the simplest case,
-        the application instatiates the MediaRecorder object, calls record() and then calls stop() or waits for the MediaStream to be ended.  The contents of the recording
-        will be made available in the platform's default encoding via the dataavailable event.  Functions are available to query
-        the platform's available set of encodings, and to select the desired ones if the author wishes.  The application can also choose
-        how much data it wants to receive at one time.  By default a Blob containing the entire recording is returned when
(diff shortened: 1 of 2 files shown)


MediaRecorder.html
```diff
@@ -340,149 +340,87 @@
...
- </section>
-
-
+  <section id="error-handling">
+    <h2>Error Handling</h2>
 
+    <section id="general-principles">
+      <h3>General Principles</h3>
 
+      <p>Errors are indicated via <a href="http://www.w3.org/TR/2012/WD-dom-20121206/#exception-domexception">DOMException</a>s raised (see [[!DOM4]]). An exception <em title="must" class="rfc2119">must</em> be thrown when the error can be detected at the time that the call is made. If recording has been started and not yet stopped when the error occurs, then after raising the error, the UA <em title="must" class="rfc2119">must</em> raise a  dataavailable event, containing any data that it has gathered, and then a stop event. The UA  <em title="may" class="rfc2119">may</em> set platform-specific limits, such those for the minimum and maximum Blob size that it will support, or the number of Tracks it will record at once.  It <em title="must" class="rfc2119">must</em> signal a fatal error if these limits are exceeded. </p>
```
foolip: DOMException is defined by WebIDL, so https://heycam.github.io/webidl/#idl-DOMException is the right reference. DOM itself (https://dom.spec.whatwg.org) references WebIDL for this.

foolip: Given that these are general principles, I would make the section non-normative and not use any RFC 2119 keywords. If there's anything here that isn't already said in one of the relevant algorithms elsewhere in the spec, that of course needs fixing first.

MediaRecorder.html
```diff
@@ -340,149 +340,87 @@
...
+          <tr>
+            <td><code>UnknownError</code></td>
+            <td>The operation failed for an unknown transient reason (e.g. out of memory). User agents SHOULD provide as much additional information as possible in the <code>message</code> attribute.</td>
+          </tr>
+          <tr>
+            <td><code>NotSupportedError</code></td>
+            <td>A modification to the <code>stream</code> has occurred that makes it impossible to continue recording. An example would be the addition of a Track while recording is occurring. User agents SHOULD provide as much additional information  as possible in the <code>message</code> attribute.</td>
+          </tr>
+          <tr>
+            <td><code>SecurityError</code></td>
```
foolip: Searching in the document (your branch opened in a browser), SecurityError is the only exception I see referenced outside of this section. This section is non-normative, so some other part of the spec needs to say when these exceptions are thrown.

foolip:
Had a look at the changes and commented in a few places. High-level, I think the biggest issues around events and exceptions are these:
1. Events are fired synchronously inside various algorithms, like the pause event in the pause() method. This is probably not intended, and not how it's implemented in Blink. Should probably say "queue a task to ..." and define what the [task source](https://html.spec.whatwg.org/#generic-task-sources) for those tasks is, and the "DOM manipulation task source" would probably be fine. (Probably some of the methods need to return and run [in parallel](https://html.spec.whatwg.org/#in-parallel).)
2. DOMError has been [removed](https://dom.spec.whatwg.org/#dom-core-changes) from DOM, and it was not an Event interface. For async errors, a simple event with name "error" plus an error attribute, or a custom error event interface like MediaRecorderErrorEvent, seems necessary.
3. The verb "raise" is used for both exceptions and events. The right terminology is e.g. "[throw](https://heycam.github.io/webidl/#dfn-throw) an InvalidStateError exception" and "[fire an event](https://dom.spec.whatwg.org/#concept-event-fire) named dataavailable".

foolip:
Who is the implementor for Firefox? Some assurance that they will adopt the changes would be good.

yell0wd0g:
Oops sorry I was addressing the first batch of changes and only saw your latest comments after uploading the added patch.

FTR, I think FF [1] has no MediaRecorderErrorEvent enumeration, which was the first pain point, and the one this patch addresses.

Also, partially related, uses a customised DOMException instead, such as foolip commented/suggested. May I suggest we limit this changes to the Exceptions, and leave the Event cleanups to another bug+patch?

[1] https://github.com/mozilla/gecko-dev/blob/master/dom/webidl/RecordErrorEvent.webidl

foolip:
I like small patches, if you want to restore MediaRecorderErrorEvent here and only deal with exceptions.

yell0wd0g:
Actually let me close this pull request and try another one from scratch :) Why?

I got confused with Exceptions and Events and mixed them both, let me follow with one single Pull Request tackling foolip@s suggestion of just dealing with DOMExceptions first, and another one with an attempt to clarify the Error Events. The first patch would be updating to what Gecko and Blink implementations do de facto, while the second one might be a tad less easy.

PR #32, opened 2015-12-07
Cleanup of DOMExceptions and/or ErrorEvents.

yell0wd0g:
A few things:
- Text reads in many places like "raise a DOM InvalidState error", whereas it actually means "throw an InvalidState DOMException". This is fixed throughout. Note that this conforms to the way both Gecko ([1] and search for "throw(" ) and Blink do it [2]. This is based on a comment by foolip@ (BIG thanks!)
- Text reads in places "raise a SecurityError event", which is corrected to "fire a SecurityError event". (foolip@ commented on this).
- the MediaRecorderErrorEvent is left untouched, to be addressed soon.
- Since there are a few DOMExceptions around, a non-normative section is added to gather them.
- Section "Error handling - General principles" is slightly reworded in its intro: "The UA must throw a DOMException (see [DOM4]) when the error can be detected at the time that the call is made. In all other cases, a MediaRecorderErrorEvent must be fired. The error name must be picked from the RecordingErrorName enums." (To be further updated soon).
- Seeing that GitHub merge is smart enough to highlight changes inside a line, I changed the HTML itself so paragraphs are coherently bundled (and not splitted in several lines etc) which seems to be made by some evil WYSIWYG. Pain now for easier review later :)

[1] https://github.com/mozilla/gecko-dev/blob/master/dom/media/MediaRecorder.cpp
[2] https://code.google.com/p/chromium/codesearch#chromium/src/third_party/WebKit/Source/modules/mediarecorder/MediaRecorder.cpp&sq=package:chromium&type=cs&l=72&q=mediarecorder.cpp%20throwDOMException

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,4 @@ MediaRecorder.diff.html
 MediaRecorder.orig.html
 MediaRecorder.orig.txt
 MediaRecorder.txt
+mediarecorder.sublime-project
--- a/MediaRecorder.html
+++ b/MediaRecorder.html
@@ -13,263 +13,162 @@
   </head>
 
 
-  <body>
-        <section id="abstract">
-                <p>This document defines a recording API for use with MediaStreams as defined in
-                        [[!GETUSERMEDIA]]</p></section>
+<body>
+ <section id="abstract">
+  <p>This document defines a recording API for use with MediaStreams as defined in [[!GETUSERMEDIA]]</p>
+ </section>
 
-  <section id="sotd">
+ <section id="sotd">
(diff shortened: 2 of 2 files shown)


yell0wd0g:
This is apparently a large code change but in reality is just a small innocent wording change (e.g. "fire and Event", "throw a DOMException" etc);  DOMExceptions is pervasive in both text and implementations, and this change just acknowledges this issue. For this reason, I'm going to commit this patch, knowing that is still not "publicised" and that we'll re-read over the same areas in the next code review, when deciding what's best for the Events. Please note if in disagree, thanks!

yell0wd0g:
And now for real :P

Issue #10, opened 2015-07-13
"MIME" as part of API name is unusual


foolip:
[`MediaRecorder.canRecordMimeType()`](http://w3c.github.io/mediacapture-record/MediaRecorder.html#widl-MediaRecorder-canRecordMimeType-DOMString-DOMString-mimeType) is in contrast to the similar APIs [`HTMLMediaElement.canPlayType()`](https://html.spec.whatwg.org/#dom-navigator-canplaytype) and [`MediaSource.isTypeSupported()`](https://w3c.github.io/media-source/#widl-MediaSource-isTypeSupported-boolean-DOMString-type).

If there isn't already content that depends on it, it would also be nice to rename [`MediaRecorder.mimeType`](http://w3c.github.io/mediacapture-record/MediaRecorder.html#widl-MediaRecorder-mimeType).

foolip:
Ping?

I would propose copying the signature of [`MediaSource`](https://w3c.github.io/media-source/#idl-def-MediaSource), which is `static boolean isTypeSupported (DOMString type)`.

Unlike for `HTMLMediaElement.canPlayType()`, there ought not be any uncertainty about whether or not a given type will result in some output. With media elements, `canPlayType("video/webm")` is supposed to return "maybe" because you don't know what the codecs will be, but when you ask to record using "video/webm", the implementation will either pick some suitable codecs that work with the container or it won't.

yell0wd0g:
I agree with the signature part: isTypeSupported().

I disagree about answering "maybe" or "probably". A number of reasons might prevent the implementation from recording a given codec+container even if it's namely supported, for example resolution when a hardware encoder is in use, or excesive frame rate, and for that reason it can not be 100% sure until the video frames start arriving. Hence the answering will be "maybe" or "" most of the times.

foolip:
What would be the distinction between "maybe" and "probably" The distinction for HTMLMediaElement is that "maybe" is returned if you just ask for a container and don't specify the codecs, because you don't know what the codecs will be. When encoding, that distinction doesn't make sense, because there are only two options:
1. A container on its own is never supported, i.e. specifying codecs is mandatory.
2. A container on its own implies some default codecs, and the return value would be the same as if the container and those default codecs had been requested.

It's still possible that encoding will fail for some reason, but the same is true of every case where HTMLMediaElement.prototype.canPlayType returns "maybe" or "probably".

yell0wd0g:
[...]

PR #33, opened 2015-12-08
s/DOMString canRecordMimeType()/bool isTypeSupported()/

yell0wd0g:
This patch proposes changing the signature of 
static DOMString canRecordMimeType(DOMString type)
with
static bool isTypeSupported(DOMString type)

in relation to issue https://github.com/w3c/mediacapture-record/issues/10. The wording is strongly inspired by [`MediaSource.isTypeSupported()`](https://w3c.github.io/media-source/#widl-MediaSource-isTypeSupported-boolean-DOMString-type) as proposed in the said issue. 

Note that this change is low risk: Gecko does not implement a static type checker [1] to the best of my knowledge and Blink implementation has only recently landed in Stable, and the use of this static method is minor, so now's the time to update it.

[1] https://github.com/mozilla/gecko-dev/blob/master/dom/webidl/MediaRecorder.webidl

### DIFF
--- a/MediaRecorder.html
+++ b/MediaRecorder.html
@@ -149,14 +149,14 @@
     </dd>
 
     <dt>static bool isTypeSupported(DOMString type)</dt>
-    <dd> Check to see whether a <code>MediaRecorder</code> can record in the specified Mime type. If true is returned from this method, it only indicates that the <code>MediaRecorder</code> implementation is capable of recording Blob objects for the specified MIME type. Recording may still fail if sufficient resources are not available to support the concrete media encoding. When this method is invoked, the user agent must run the following steps:
+    <dd> Check to see whether a <code>MediaRecorder</code> can record in the specified Mime type. If true is returned from this method, it only indicates that the <code>MediaRecorder</code> implementation is capable of recording Blob objects for the specified MIME type. Recording may still fail if sufficient resources are not available to support the concrete media encoding. When this method is invoked, the User Agent must run the following steps:
      <ol class="method-algorithm">
       <li>If <code>type</code> is an empty string, then return false.</li>
       <li>If <code>type</code> does not contain a valid MIME type string, then return false.</li>
-      <li>If <code>type</code> contains a media type or media subtype that the MediaSource does not support, then return false.</li>
+      <li>If <code>type</code> contains a media type or media subtype that the MediaRecorder does not support, then return false.</li>
       <li>If <code>type</code> contains a media container that the MediaSource does not support, then return false.</li>
       <li>If <code>type</code> contains a codec that the MediaSource does not support, then return false.</li>
-      <li>If the MediaSource does not support the specified combination of media type, media subtype, and codecs then return false.</li>
+      <li>If the MediaRecorder does not support the specified combination of media type/subtype, codecs and container then return false.</li>
       <li>Return true.</li>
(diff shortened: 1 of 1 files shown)


foolip:
> Blink implementation has only recently landed in Stable

To clarify, it is not yet enabled by default, it's still behind an experimental flag.

MediaRecorder.html
```diff
@@ -148,12 +148,21 @@
...
-    <dd><p>This method <em title="must" class="rfc2119">must</em> return the empty string if <code>mimeType</code> is a type that the user agent knows it cannot record; it <em title="must" class="rfc2119">must</em> return "probably" if the user agent is confident that <code>mimeType</code> represents a type that it can record; and it <em title="must" class="rfc2119">must</em> return "maybe" otherwise. Implementors are encouraged to return "maybe" unless the type can be confidently established as being supported or not.</p>
+    <dt>static bool isTypeSupported(DOMString type)</dt>
+    <dd> Check to see whether a <code>MediaRecorder</code> can record in the specified Mime type. If true is returned from this method, it only indicates that the <code>MediaRecorder</code> implementation is capable of recording Blob objects for the specified MIME type. Recording may still fail if sufficient resources are not available to support the concrete media encoding. When this method is invoked, the user agent must run the following steps:
+     <ol class="method-algorithm">
+      <li>If <code>type</code> is an empty string, then return false.</li>
```
foolip, line 154: Unless an empty string is a valid MIME type string, then this step is redundant.

MediaRecorder.html
```diff
@@ -148,12 +148,21 @@
...
+      <li>If <code>type</code> does not contain a valid MIME type string, then return false.</li>
+      <li>If <code>type</code> contains a media type or media subtype that the MediaSource does not support, then return false.</li>
+      <li>If <code>type</code> contains a media container that the MediaSource does not support, then return false.</li>
+      <li>If <code>type</code> contains a codec that the MediaSource does not support, then return false.</li>
```
foolip, line 158: It would also be good to require that the combination of container is codec is required, so that "video/webm;codecs=theora" would return false even though both the WebM container and Theora codec is supported, just not together.

foolip:
Changes look great, left a few comments.

MediaRecorder.html
yell0wd0g, line 154: I believe it was there indeed to address the uncertainty of empty-mime-type-is-good-or-not question. I'm thinking about methods buried deep into a JS hierarchy...

MediaRecorder.html
yell0wd0g, line 158: Yeah, the next line was there to address that, I rewrote it though for more clarification and should read sth like: 

> If the MediaRecorder does not support the specified combination of media type/subtype, codecs and container then return false.

MediaRecorder.html
foolip, line 154: If the uncertainty is on the part of the implementor, you could add a non-normative note to say that the empty string is not a valid MIME type, or similar.

MediaRecorder.html
foolip, line 158: Oh, I actually missed that. It seems like that line enough would be enough as the definition, since it encompasses all other problems. Editorial only, though.

mreavy:
Already landed, but LGTM

PR #37, opened 2015-12-09
typos pointed out by foolip@ on https://github.com/w3c/mediacapture-record/pull/25

yell0wd0g:


MediaRecorder.html
```diff
@@ -182,11 +182,11 @@
 <section id="blob-event">
     <h2>Blob Event</h2>
     <dl title='interface BlobEvent : Event' class='idl'>
-        <dt>Constructor(DOMString type, optional BlobEventInit eventInit)</dt>
+        <dt>Constructor(DOMString type, BlobEventInit eventInit)</dt>
```
foolip, line 185: The data member of BlobEventInit also needs the required keyword. Without it, this is actually invalid per WebIDL, which [says](http://heycam.github.io/webidl/#idl-operations) "If the type of an argument is a dictionary type or a union type that has a dictionary type as one of its flattened member types, and that dictionary type and its ancestors have no required members, and the argument is either the final argument or is followed only by optional arguments, then the argument must be specified as optional. Such arguments are always considered to have a default value of an empty dictionary, unless otherwise specified."

If you're changing this again, I would also suggest s/eventInit/eventInitDict/ for consistency with other specs.

yell0wd0g, line 185: Doing it in https://github.com/w3c/mediacapture-record/pull/38 :)

PR #38, opened 2015-12-09
Add required keyword, and s/eventInit/eventInitDict/

yell0wd0g:
Bc of https://github.com/w3c/mediacapture-record/pull/37/files#r47140210

### DIFF
--- a/MediaRecorder.html
+++ b/MediaRecorder.html
@@ -182,11 +182,11 @@
 <section id="blob-event">
     <h2>Blob Event</h2>
     <dl title='interface BlobEvent : Event' class='idl'>
-        <dt>Constructor(DOMString type, BlobEventInit eventInit)</dt>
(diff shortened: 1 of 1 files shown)


MediaRecorder.html
```diff
@@ -182,11 +182,11 @@
 <section id="blob-event">
     <h2>Blob Event</h2>
     <dl title='interface BlobEvent : Event' class='idl'>
-        <dt>Constructor(DOMString type, BlobEventInit eventInit)</dt>
+        <dt>Constructor(DOMString type, required BlobEventInit eventInitDict)</dt>
```
foolip, line 185: Actually, you can't use the required keyword here, all arguments are required unless they have the optional keyword. The required keyword is only for dictionary members, as you've done below.

yell0wd0g, line 185: I just found out too. Moving the |required| to BlobEventInit.data in https://github.com/w3c/mediacapture-record/pull/39

PR #39, opened 2015-12-09
corrected: |required| goes to BlobEventInit.data

yell0wd0g:
Re. comment in https://github.com/w3c/mediacapture-record/pull/38/files#r47154100


foolip:
:+1:

Issue #4, opened 2015-02-23
MediaRecorder needs to define effect of adding / removing tracks in its input MediaStream


ghost:
Shijun Sun 2014-05-20 20:13:29 UTC

The "stream" attribute on the MediaRecorder is readonly, but it seems the spec does not prevent the app from dynamically adding/removing tracks from the stream object.

If the intention of defining the attribute as readonly is to block dynamic changes to the stream object, it will be better to define a getStream() method to return a copy of the internal MediaStream object.

Comment 1 Adam Bergkvist 2014-05-21 13:01:40 UTC

The readonly only prevents the script from assigning the attribute to an entirely new MediaStream. Returning a copy with getStream() wouldn't help since you can still have the reference to the stream you passed in to the MediaRecorder() constructor. To prevent this, we would have to have a "readonly mode" for the stream. I'm not sure we want/need that.

Comment 2 Martin Thomson 2014-05-21 15:24:12 UTC

The point here is that we need some sort of rules for handling the cases where tracks are added and removed from a stream that is in active use.

Comment 3 Harald Alvestrand 2014-06-13 08:41:18 UTC

Changing the subject of the bug to say what it's about (based on discussion).

The alternative would be to close as WONTFIX/WAI.

mhofman:
Somebody raised a use case on the Chrome launch bug to be able to switch video tracks in a single recording: https://code.google.com/p/chromium/issues/detail?id=261321#c62

I believe we should support the case where a stream have its tracks changing.

From what I remember of our discussion at TPAC, this actually might happen without the app's action when using `captureStream` from a video element that has its played back tracks changing.

yell0wd0g:
Chrome's CL [1] brought the question in the Title. Previous to https://github.com/w3c/mediacapture-record/pull/46, it seemed that a Track addition to/removal from a MediaStream being recorded should signal some error condition _and_ stop the recording. The argument in said CL is that the JS should decide what to do, since the addition/removal might be desired, and even if it was not, it's unclear what the Application might deem appropriate: start recording the new track seamlessly, stop the recording altogether, stop and restart with the new Track... 

I propose not being dogmatic, signal to the App the condition and let it decide what to do -- in that line of reasoning, perhaps firing an Error when a Track is added/removed might also not be a good idea. What about a new `mediastreammodified` Event?

@mreavy I don't see any firing/throwing being implemented in Gecko now [2], correct? 

[1] https://crrev.com/1578973002
[...]

PR #46, opened 2016-01-20
Removed UnknownError Exception; added ErrorEvent description.

yell0wd0g:
Neither Gecko [1] nor Chrome [2] throw an UnknownError Exception. Moreover, an Exception is supposed to be thrown for a synchronous Event, whereas the detailed "out of memory" or "a track has been added to/removed from the recorded MediaStream" is asynchronous. This PR moves these types of Errors to the Event Section.

[1] https://github.com/mozilla/gecko-dev/blob/master/dom/media/MediaRecorder.cpp
[2] https://code.google.com/p/chromium/codesearch#chromium/src/third_party/WebKit/Source/modules/mediarecorder/MediaRecorder.cpp&sq=package:chromium&type=cs&q=mediarecorder.cpp

### DIFF
--- a/MediaRecorder.html
+++ b/MediaRecorder.html
@@ -302,11 +302,6 @@
           <td> An operation was called on an object on which it is not allowed or at a time when it is not allowed, or if a request is made on a source object that has been deleted or removed. </td>
         </tr>
         <tr>
-          <td><code>UnknownError</code></td>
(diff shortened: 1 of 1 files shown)


yell0wd0g:
This also touches issue https://github.com/w3c/mediacapture-record/issues/4.

mhofman:
I'm really concerned that the wording change has impact outside of just changing the way errors are notified. Dropping the words `that makes it impossible to continue recording` touches on issue #4.
Before, it was implying that some modifications to the stream wouldn't cause Errors if the user-agent could deal with them. Now it reads like any modifications should result in an error.

yell0wd0g:
@mhofman, removal unintended, adding it again.

Issue #43, opened 2016-01-18
Provide Surface API for reading back the configured encoding bitrate(s)


yell0wd0g:
Section [MediaRecorder Properties](https://rawgit.com/w3c/mediacapture-record/master/MediaRecorder.html#properties) says:

> [TO DO: Provide API surface to set and read these. When that is done, this section will be removed.]

this is the issue for that TODO, for the bitrate options concretely.

(It's basically a follow up of @mreavy "improve bitrate control in MediaRecorder" #20 - discussed with @tdaede).

mreavy:
That looks good, except that the three bitrate properties should all use the same or similar language to say that the target bitrate isn't guaranteed.  I'll close #20

yell0wd0g:
@mreavy, re. your comment, from https://rawgit.com/w3c/mediacapture-record/master/MediaRecorder.html#dictionary-mediarecorderoptions-members, 
- `audioBitsPerSecond`:  `This is a hint for the encoder and the value might only be achieved over a long time.`
- `videoBitsPerSecond` : `This is a hint for the encoder and the value might be surpassed or not achieved.`
- `bitsPerSecond`: `This parameter is a hint for the encoder(s) and the total value might be surpassed or not achieved at all.`

mreavy:
@miguelao :I'm sorry, I don't see a question here - those are the descriptions that need to be synchronized, yes.

yell0wd0g:
Oh I see, you meant that the three should say a similar thing, gotcha.

yell0wd0g:
Closed as per #47

PR #47, opened 2016-01-20
Added {audio,video}BitsPerSecond read-only attributes.

yell0wd0g:
Also Removed some stale todo-properties entries and reflowed html in that area.

Re. https://github.com/w3c/mediacapture-record/issues/43

### DIFF
--- a/MediaRecorder.html
+++ b/MediaRecorder.html
@@ -85,6 +85,12 @@
     <dt>attribute boolean ignoreMutedMedia</dt>
     <dd>If this attribute is set to <code>true</code>, the MediaRecorder will not record anything when the input Media Stream is muted.  If this attribute is <code>false</code>, the MediaRecorder will record silence (for audio) and black frames (for video) when the input MediaStream is muted.  When the MediaRecorder is created, the UA <em title="must" class="rfc2119">must</em> set this attribute to <code>false</code>.</dd>
 
+    <dt>readonly attribute unsigned long videoBitsPerSecond</dt>
+    <dd>The value of the Video encoding target bit rate that was passed to the Platform. If the user has not specified any, the Platform implementation is free to choose this value and it <em title="must" class="rfc2119">must</em> read zero.</dd>
+
+    <dt>readonly attribute unsigned long audioBitsPerSecond</dt>
+    <dd>The value of the Audio encoding target bit rate that was passed to the Platform. If the user has not specified any, the Platform implementation is free to choose this value and it <em title="must" class="rfc2119">must</em> read zero.</dd>
+
     <dt>void start()</dt>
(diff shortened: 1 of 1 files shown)


yell0wd0g:
@foolip, @mreavy can you please take a look?thx

yell0wd0g:
I thought about publishing the consolidated audio and video bit rates. The application knows what was written in the `options`, but due to clamping or calculations based on `bitsPerSecond`, the individual values could be different. Perhaps an app left the decision on the actual values to the UA and is interested in retrieving the configured ones. For all that, I don't think `bitsPerSecond` should be there. WDYT?

yell0wd0g:
@mreavy, I took the chance of this pull request to homogeneise the wording you mentioned in https://github.com/w3c/mediacapture-record/issues/43.

MediaRecorder.html
```diff
@@ -215,52 +221,25 @@
...
+      <th>Property Name</th>
+      <th>Values</th>
+      <th>Notes</th>
+    </tr>
+  </thead>
+  <tbody>
+    <tr id="def-property-MimeType">
+      <td>MimeType</td>
+      <td><code>DOMString</code></td>
+      <td> The MIME type that was be selected as the container for recording. This entry includes all the parameters to the base MimeType. The UA should be able to play back any of the MIME types it supports for recording. For example, it should be able to display a video recording in the HTML &lt;video&gt; tag. The default value for this property is platform-specific.</td>
```
foolip: This says "was be selected" and both "MIME type" and "MimeType".

foolip: I see this was just moved and reformatted in this commit, but might want to fix anyway :)

foolip:
I don't have anything to object here, as long as `audioBitsPerSecond` and `videoBitsPerSecond` are only added to the options dictionary, that's just matching Gecko and not opening up any new issues I think. Aggregate bitrates or exposing the actually chosen target bitrate might be trickier.

mreavy:
@miguelao The wording looks good, thanks.

MediaRecorder.html
yell0wd0g: done both, thanks!

yell0wd0g:
Sorry I re-read what the original patch was stating:

>  If the user has not specified any, the Platform implementation is free to choose this value and it <em title="must" class="rfc2119">must</em> read zero.</dd>

And I realised that my comments where not making much sense. I removed that line and substituted with 

>  or the calculated one if the user has specified <code>bitsPerSecond</code>.

And my comments apply to such writing. Apologies :P

yell0wd0g:
ping @foolip can you check the latest patch?

foolip:
Changes look good, thanks!

PR #50, opened 2016-02-03
Remove MediaRecorder Properties Section, which was a TODO. 

yell0wd0g:
Also added a tiny line of comments in `{audio,video}BitsPerSecond` `readonly attribute` to detail as to why is interesting to have this accessors (it's interesting because what the user provides can be truncated, rounded, and in some cases, such as when `bitsPerSecond` is provided, the UA can allocate the budget to audio and video as it sees fit, and JS might very likely be interested in these numbers).


yell0wd0g:
Going ahead and merging since these are just editorial comments.

Issue #59, opened 2016-05-31
Provide an option to enumerate all supported media types


sole:
Right now the method `isTypeSupported` provides _some amount_ of media support detection, but developers might not be able to come up with all the possible combinations of codec and rate (e.g. `aType=video/webm; codecs="vp9"` vs `aType=video/mp4; codecs="avc1.4d4015"; width=426; height=240`, etc

Could it be possible to provide a method that enumerates the available options, so that developers can provide that to the users?

This is an often requested feature when talking to developers about `MediaRecorder`

yell0wd0g:
I think the original idea might have been to let the implementation decide which codec/container was best suited and hide all those details from the user. In actuality, however, the developer will very likely need to know/specify such info in order to know the file type generated and, moreover, different implementations of MR today support different configurations. With all that, enumerating the capabilities might be a very good idea. 

I would also propose that they come ordered by the implementation by weak-preference order, IOW, the first ones in the list are the preferred ones, e.g. because there is support for accelerated encoding. 

Something in the lines of `static sequence<DOMString> getSupportedTypes ();`, with each DOMString being a `valid MIME type [HTML5]`.

Currently the API provides a `static bool isTypeSupported (DOMString type);`, I would say we should keep it for backwards compatibility and convenience.

sole:
Yes! `getSupportedTypes` would be very nice - same as keeping `isTypeSupported` because it's more succinct than searching in the list returned by `getSupportedTypes`. Convenience!

fluffy:
Often people want to be able to do something with the recording - thus why they made it. So they might need some control over what the format is. The browser can not figure out what is best with no hints.

yell0wd0g:
@mreavy -- would you know someone in FF that could take a look at the PR #65 and have an opinion on it?  I think this proposal has developer support and does not increase the fingerprinting surface, as I've argued in the said PR?

yell0wd0g:
I think it's best to continue the conversation at https://github.com/WICG/media-capabilities/issues/28.

npdoty:
WICG/media-capabilities#28 has been closed, but I don't see an explanation of the conclusion. Is this issue still relevant on this spec or is the discussion to be moved elsewhere?

[...]

PR #64, opened 2016-07-21
Micro draft of getSupportedTypes()

yell0wd0g:
@foolip can you take a look at the proposed method and let me know what you think, if it's pro forma etc?

Also, can a function be static and return a Promise<> ? :)

Relates to #59.


foolip:
Static methods can return promises, here are two examples I found by grepping Blink:
https://notifications.spec.whatwg.org/#dom-notification-requestpermission
https://w3c.github.io/webrtc-pc/#sec.cert-mgmt

However, in this case it seems odd that you can synchronously ask if a type is supported, but asking for the supported types is async. What kind of implementation would actually have this restriction? In other words, can't getSupportedTypes also be sync?

yell0wd0g:
`getSupportedTypes()` might need to check which codecs are/can be e.g. hardware accelerated, whereas `isTypeSupported()` can answer questions more easily (checking on a list, essentially). 

An example: H264 is supported in Chrome/Blink in all cases in Sw and, for certain platforms and hardware configurations, by a encode accelerator. `isTypeSupported()` knows is supported, whereas `getSupportedTypes()` might need to dig into lower levels of the implementation for to bump up this format in the list.

MediaRecorder.html
```diff
@@ -442,6 +443,20 @@
...
+
+
+          <dt><dfn><code>getSupportedTypes</code></dfn>, static</dt>
+          <dd>
+            Provides a list of the <a href="https://tools.ietf.org/html/rfc2046">
+            MIME types</a> supported by the platform. The returned list is ordered
+            so that MIME types closer to the beginning have a larger probability
+            of being supported by encode accelerators.
+            <div>
+              <em>Return type:</em> <code>sequence&lt;DOMString&gt;</code>
```
foolip: It returns a promise, not a sequence directly. Wording more in the style of https://notifications.spec.whatwg.org/#dom-notification-requestpermission would be nice :)

foolip:
Ah, the order tells you additional information, I didn't read very carefully. It makes sense, concept LGTM.

foolip:
BTW, do you have someone to more carefully review spec changes, or am I it here?

yell0wd0g:
Too many rebases :)
Will close this PR and open another fresh one.

PR #65, opened 2016-07-28
getSupportedTypes()

yell0wd0g:
Addresses #59, and is a refry of #64.

@foolip can you take a look?

### DIFF
--- a/MediaRecorder.html
+++ b/MediaRecorder.html
@@ -462,8 +462,9 @@ interface MediaRecorder : EventTarget {
              </a> [[!HTML5]]. The codecs parameter [[!RFC6381]], if necessary to
              specify exactly how the resource is encoded, SHOULD be added
              to <code>type</code>. If <var>type</var> can be expressed in several
-             equivalent ways (e.g. with or without quotes), the shortest form
-             and lowercase letters SHOULD be preferred.
+             equivalent ways (e.g. with or without quotes or whitespaces), the
+             one using the smallest amount of lowercase characters MUST be
+             preferred.
 
              </a>.
              <var>resultList</var> MAY be ordered by the UA so that MIME types
--- a/MediaRecorder.html
+++ b/MediaRecorder.html
@@ -458,12 +458,12 @@ interface MediaRecorder : EventTarget {
              multiplexing (i.e. <code>isSuportedType(<var>type</var>)</code>
              is true), add <var>type</var> to <var>resultList</var>.
 
-             <var>type</var> must be a <a def-id="mime-types">valid MIME type
+             <var>type</var> MUST be a <a def-id="mime-types">valid MIME type
              </a> [[!HTML5]]. The codecs parameter [[!RFC6381]], if necessary to
-             specify exactly how the resource is encoded, should be added
+             specify exactly how the resource is encoded, SHOULD be added
              to <code>type</code>. If <var>type</var> can be expressed in several
              equivalent ways (e.g. with or without quotes), the shortest form
-             and lowercase letters should be preferred.
+             and lowercase letters SHOULD be preferred.
 
              </a>.
              <var>resultList</var> MAY be ordered by the UA so that MIME types
--- a/MediaRecorder.html
+++ b/MediaRecorder.html
@@ -80,6 +80,7 @@ interface MediaRecorder : EventTarget {
     void           resume ();
     void           requestData ();
     static boolean isTypeSupported (DOMString type);
+    static Promise&lt;sequence&lt;DOMString&gt;&gt; getSupportedTypes ();
 };</pre>
       <section>
         <h2>Constructors</h2>
@@ -442,6 +443,38 @@ interface MediaRecorder : EventTarget {
               <em>Return type:</em> <code>boolean</code>
             </div>
           </dd>
+
+          <dt><dfn><code>getSupportedTypes</code></dfn>, static</dt>
+          <dd>
+            When <code>getSupportedTypes()</code> method is invoked, the UA MUST
+            queue a task, using the DOM manipulation task source, that runs the
+            following steps:
+            <ol>
+            <li>Let <var>p</var> be a promise.</li>
+            <li>Run the following steps in parallel:
+             <ol>
+             <li>Let <var>resultList</var> be an empty list.
+             <li>For each <a href="https://tools.ietf.org/html/rfc2046">MIME
+             type</a> <var>type</var> that is supported for encoding and/or
+             multiplexing (i.e. <code>isSuportedType(<var>mimeType</var>)</code>
+             is true), add <var>type</var> to <var>resultList</var>.
+             </li>
+             <li>Resolve <var>p</var> with <var>resultList</var>.</li>
+             </ol>
+             </li>
+            <li> Return <var>p</var></li>
+            </ol>
+            <div>
+              <p class="note"> <var>resultList</var> MAY be ordered by the UA so
+              that MIME types closer to the beginning have a larger probability
+              of being supported by encode accelerators.</p>
+            </div>
+            <div>
+              <em>Return type:</em>
+              <code>Promise&lt;sequence&lt;DOMString&gt;&gt;</code>
+            </div>
+          </dd>
+
         </dl>
       </section>
     </div>
@@ -832,6 +865,7 @@ interface ErrorEvent : Event {
       <li>Added read-only attributes videoBitsPerSecond and
       audioBitsPerSecond.</li>
       <li>Removed MediaRecorder Properties Section, which was a TODO.</li>
+      <li>Added method <code>getSupportedTypes()<code></li>
     </ol>
   </section>
 </body>
--- a/MediaRecorder.html
+++ b/MediaRecorder.html
@@ -454,21 +454,23 @@ interface MediaRecorder : EventTarget {
             <li>Run the following steps in parallel:
              <ol>
              <li>Let <var>resultList</var> be an empty list.
-             <li>For each <a href="https://tools.ietf.org/html/rfc2046">MIME
-             type</a> <var>type</var> that is supported for encoding and/or
+             <li>For each <dfn>minimal</dfn> <a href="https://tools.ietf.org/html/rfc2046">
+             MIME type</a> <var>type</var> that is supported for encoding and/or
              multiplexing (i.e. <code>isSuportedType(<var>mimeType</var>)</code>
              is true), add <var>type</var> to <var>resultList</var>.
+             A <dfn>minimal</dfn> <a href="https://tools.ietf.org/html/rfc2046">
+             MIME type</a> is composed, for the scope of this method, following
+             the rules of <a href="https://tools.ietf.org/html/rfc6381">rfc6381
+             </a>.
+             <var>resultList</var> MAY be ordered by the UA so that MIME types
+             closer to the beginning are preferred, for example due to larger
+             probability of being supported by encode accelerators.
              </li>
              <li>Resolve <var>p</var> with <var>resultList</var>.</li>
              </ol>
              </li>
-            <li> Return <var>p</var></li>
+            <li>Return <var>p</var>.</li>
             </ol>
-            <div>
-              <p class="note"> <var>resultList</var> MAY be ordered by the UA so
-              that MIME types closer to the beginning have a larger probability
-              of being supported by encode accelerators.</p>
-            </div>
             <div>
               <em>Return type:</em>
               <code>Promise&lt;sequence&lt;DOMString&gt;&gt;</code>
--- a/MediaRecorder.html
+++ b/MediaRecorder.html
@@ -463,7 +463,8 @@ interface MediaRecorder : EventTarget {
              specify exactly how the resource is encoded, SHOULD be added
              to <code>type</code>. If <var>type</var> can be expressed in several
              equivalent ways (e.g. with or without quotes or whitespaces), the
-             one using the smallest amount of lowercase characters MUST be
+             shortest one, employing the least amount of whitespaces, and
+             with the largest amount of lowercase characters MUST be
              preferred.
 
              </a>.
--- a/MediaRecorder.html
+++ b/MediaRecorder.html
@@ -446,20 +446,23 @@ interface MediaRecorder : EventTarget {
 
           <dt><dfn><code>getSupportedTypes</code></dfn>, static</dt>
           <dd>
-            When <code>getSupportedTypes()</code> method is invoked, run the
-            following steps:
+            When this method is invoked, the User Agent must run the following
+            steps:
             <ol>
             <li>Let <var>p</var> be a promise.</li>
             <li>Run the following steps in parallel:
              <ol>
              <li>Let <var>resultList</var> be an empty list.
-             <li>For each <dfn>minimal</dfn> <a href="https://tools.ietf.org/html/rfc2046">
-             MIME type</a> <var>type</var> that is supported for encoding and/or
-             multiplexing (i.e. <code>isSuportedType(<var>mimeType</var>)</code>
+             <li>For each <a href="https://tools.ietf.org/html/rfc2046">MIME
+             type</a> <var>type</var> that is supported for encoding and/or
+             multiplexing (i.e. <code>isSuportedType(<var>type</var>)</code>
              is true), add <var>type</var> to <var>resultList</var>.
-             A <dfn>minimal</dfn> <a href="https://tools.ietf.org/html/rfc2046">
-             MIME type</a> is composed, for the scope of this method, following
-             the rules of <a href="https://tools.ietf.org/html/rfc6381">rfc6381
+
+             <var>type</var> must be a <a def-id="mime-types">valid MIME type
+             </a> [[!HTML5]]. The codecs parameter [[!RFC6381]], if necessary to
+             specify exactly how the resource is encoded, should be added
+             to <code>type</code>.
+
              </a>.
              <var>resultList</var> MAY be ordered by the UA so that MIME types
              closer to the beginning are preferred, for example due to larger
--- a/MediaRecorder.html
+++ b/MediaRecorder.html
@@ -461,11 +461,11 @@ interface MediaRecorder : EventTarget {
              <var>type</var> MUST be a <a def-id="mime-types">valid MIME type
              </a> [[!HTML5]]. The codecs parameter [[!RFC6381]], if necessary to
              specify exactly how the resource is encoded, SHOULD be added
-             to <code>type</code>. If <var>type</var> can be expressed in several
-             equivalent ways (e.g. with or without quotes or whitespaces), the
-             shortest one, employing the least amount of whitespaces, and
-             with the largest amount of lowercase characters MUST be
-             preferred.
+             to <code>type</code>.
+             If <var>type</var> can be expressed in several equivalent ways
+             (e.g. with or without quotes and whitespace, or different
+             capitalization), the shortest form with the largest number of
+             lowercase characters MUST be used.
 
              </a>.
              <var>resultList</var> MAY be ordered by the UA so that MIME types
--- a/MediaRecorder.html
+++ b/MediaRecorder.html
@@ -446,8 +446,7 @@ interface MediaRecorder : EventTarget {
 
           <dt><dfn><code>getSupportedTypes</code></dfn>, static</dt>
           <dd>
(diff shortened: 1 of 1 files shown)


MediaRecorder.html
```diff
@@ -442,6 +443,38 @@
               <em>Return type:</em> <code>boolean</code>
             </div>
           </dd>
+
+          <dt><dfn><code>getSupportedTypes</code></dfn>, static</dt>
+          <dd>
+            When <code>getSupportedTypes()</code> method is invoked, the UA MUST
+            queue a task, using the DOM manipulation task source, that runs the
```
foolip: Why does it need to queue a task at the beginning? Tests to ensure that promise isn't resolved too fast would be kind of fun (really!) but I haven't seen this pattern before.

MediaRecorder.html
```diff
@@ -442,6 +443,38 @@
...
+            following steps:
+            <ol>
+            <li>Let <var>p</var> be a promise.</li>
+            <li>Run the following steps in parallel:
+             <ol>
+             <li>Let <var>resultList</var> be an empty list.
+             <li>For each <a href="https://tools.ietf.org/html/rfc2046">MIME
```
foolip, line 456: This needs to say something about normalization, presumably there are an infinite number of strings for which `isSupportedType(mimeType)` returns true because of whitespace?

Also needs to say something about codecs, for example should both `video/webm` and `video/webm; codecs=vorbis,vp8` be returned, or which set of string?

MediaRecorder.html
```diff
@@ -442,6 +443,38 @@
...
+             type</a> <var>type</var> that is supported for encoding and/or
+             multiplexing (i.e. <code>isSuportedType(<var>mimeType</var>)</code>
+             is true), add <var>type</var> to <var>resultList</var>.
+             </li>
+             <li>Resolve <var>p</var> with <var>resultList</var>.</li>
+             </ol>
+             </li>
+            <li> Return <var>p</var></li>
```
foolip: Extra whitespace before Return and missing period.

MediaRecorder.html
```diff
@@ -442,6 +443,38 @@
...
+            </ol>
+            <div>
+              <p class="note"> <var>resultList</var> MAY be ordered by the UA so
+              that MIME types closer to the beginning have a larger probability
+              of being supported by encode accelerators.</p>
```
foolip: Is hardware support the only reason one might promote something to the beginning, and what if there are multiple codecs with hardware support? Can you bake this advice into the normative part of  the algorithm (keyword MAY is normative) and say that `resultList` must be ordered such that the formats that the UA would prefer the web application to use are at the front, for whatever reason?

foolip:
Dug this out of my inbox, looks entirely sane. Where are the other implementors on this? It'd be very good if they weighed in. Also wonderful would be an explainer for https://github.com/w3c/mediacapture-record which shows how you might use this (and the other parts of the API).

MediaRecorder.html
yell0wd0g: Done.

MediaRecorder.html
yell0wd0g: Done.

MediaRecorder.html
yell0wd0g, line 456: Addressed, I think. RFC6381 talks about how to define these combinations. I think that returning both "video/webm" and "video/webm;codecs=vp8" should be OK.

MediaRecorder.html
yell0wd0g: I have no clue, I think I siphoned this phrasing off from someplace else describing how to express correctly Promise-based methods, but is possible that I frankensteneised it in flight. How would you phrase it?

foolip: I'd just say "When the <code>getSupportedTypes()</code> method is invoked, run the following steps", since all it does (synchronously) is to create and return a promise.

MediaRecorder.html
foolip: How about normalization? Can any of the RFCs really say whether there should be a space after the semicolon, or when the codecs list should be quoted?

MediaRecorder.html

MediaRecorder.html
yell0wd0g: RFC2045 says a couple of things:

> [...] the remainder of the header field is simply a set of parameters, specified in an attribute=value notation.  The ordering of parameters is not significant.

Also, 

> [...] MIME implementations must ignore any parameters whose names they do not recognize.

Note that `video/webm` and `audio/webm` are not officially assigned by IANA ([here](http://www.iana.org/assignments/media-types/media-types.xhtml#video) is the list), so the only definition comes, informally, from the [webm web page](http://www.webmproject.org/docs/container/).

foolip: Hmm, so from that (or any other spec) would you say that "video/webm;codecs=vp8" or "video/webm; codecs=vp8" is the expected string from this API? If it is actually defined, it would still be good to call out in a non-normative note what is expected and why, so that there can't be any divergence on silly things like whitespace, capitalization or quotes.

MediaRecorder.html
```diff
@@ -446,8 +446,7 @@
 
           <dt><dfn><code>getSupportedTypes</code></dfn>, static</dt>
           <dd>
-            When <code>getSupportedTypes()</code> method is invoked, the UA MUST
-            queue a task, using the DOM manipulation task source, that runs the
+            When <code>getSupportedTypes()</code> method is invoked, run the
```
foolip: Can you add "the" here? Or perhaps "this method" which is how isTypeSupported puts it.

foolip:
Took another look, still looks sane, but would benefit from other vendors weighing in.

MediaRecorder.html
yell0wd0g: Asked around internally and I took inspiration on the [Encripted Media Extensions Spec](https://w3c.github.io/encrypted-media/#dom-mediakeysystemmediacapability-contenttype), essentially referring to `valid MIME type` as per HTML5 def, and RFC6381 for the parameters/codecs.

yell0wd0g:
Took in @foolip comments. 

@martinthomson can you comment on this PR, or would you know someone from Gecko with some opinion they'd like to share?

MediaRecorder.html
foolip: The [valid MIME type](https://html.spec.whatwg.org/multipage/infrastructure.html#valid-mime-type) concept points to  https://tools.ietf.org/html/rfc7231#section-3.1.1.1 and crucially the syntax there has optional whitespace (OWS) and other optional quoting. There's an example given:

> The quoted and unquoted values are equivalent.  For example, the following  examples are all equivalent, but the first is preferred for consistency:
> text/html;charset=utf-8
> text/html;charset=UTF-8
> Text/HTML;Charset="utf-8"
> text/html; charset="utf-8"

So, I think this spec (and EME) needs to constrain the string to the point where no variation is possible, for example by saying that it should be the shortest possible valid MIME type, where every character that can be lowercase is lowercase.

yell0wd0g: You're right. Wrote some more literature on this.

MediaRecorder.html
```diff
@@ -442,6 +443,44 @@
...
+             <li>For each <a href="https://tools.ietf.org/html/rfc2046">MIME
+             type</a> <var>type</var> that is supported for encoding and/or
+             multiplexing (i.e. <code>isSuportedType(<var>type</var>)</code>
+             is true), add <var>type</var> to <var>resultList</var>.
+
+             <var>type</var> MUST be a <a def-id="mime-types">valid MIME type
+             </a> [[!HTML5]]. The codecs parameter [[!RFC6381]], if necessary to
+             specify exactly how the resource is encoded, SHOULD be added
+             to <code>type</code>. If <var>type</var> can be expressed in several
+             equivalent ways (e.g. with or without quotes), the shortest form
```
foolip: Also mention whitespace here?

MediaRecorder.html
```diff
@@ -442,6 +443,44 @@
...
+             and lowercase letters SHOULD be preferred.
```
foolip: MUST?

MediaRecorder.html
yell0wd0g: I'd argue that "the shortest" imply having the least amount of whitespaces, but in the interest of clarity, I rewrote it.

MediaRecorder.html
yell0wd0g: Done

MediaRecorder.html
```diff
@@ -462,8 +462,9 @@
              </a> [[!HTML5]]. The codecs parameter [[!RFC6381]], if necessary to
              specify exactly how the resource is encoded, SHOULD be added
              to <code>type</code>. If <var>type</var> can be expressed in several
-             equivalent ways (e.g. with or without quotes), the shortest form
-             and lowercase letters SHOULD be preferred.
+             equivalent ways (e.g. with or without quotes or whitespaces), the
+             one using the smallest amount of lowercase characters MUST be
```
foolip: The "shortest form" bit was important. From the new wording "video/WebM" is better than "video/webm" because in fact it has fewer lowercase characters :)

yell0wd0g: Heh I have written all this so many times that I just don't know what I'm typing. I rewrote again this section but, if you still have comments (all welcome), may I ask you suggest an example wording? ;)

martinthomson:
Why not use [canPlayType](https://html.spec.whatwg.org/multipage/embedded-content.html#dom-navigator-canplaytype) (maybe renamed as `canRecordType`) instead? Then you can reference the rules in the HTML spec.

MediaRecorder.html
```diff
@@ -442,6 +443,46 @@
...
+             type</a> <var>type</var> that is supported for encoding and/or
+             multiplexing (i.e. <code>isSuportedType(<var>type</var>)</code>
+             is true), add <var>type</var> to <var>resultList</var>.
+
+             <var>type</var> MUST be a <a def-id="mime-types">valid MIME type
+             </a> [[!HTML5]]. The codecs parameter [[!RFC6381]], if necessary to
+             specify exactly how the resource is encoded, SHOULD be added
+             to <code>type</code>. If <var>type</var> can be expressed in several
+             equivalent ways (e.g. with or without quotes or whitespaces), the
+             shortest one, employing the least amount of whitespaces, and
```
foolip: Sorry for making this so much fun for you :) Since you've given me permission to tweak, I'd like "If <var>type</var> can be expressed in several equivalent ways (e.g. with or without quotes and whitespace, or different capitalization), the shortest form with the largest number of lowercase characters MUST be used."

foolip:
@martinthomson, canPlayType doesn't return a string, so HTML doesn't ever require UAs to serialize these strings, only to parse them.

yell0wd0g:
Ping @foolip, @martinthomson.

martinthomson:
@foolip that's why it is superior.

foolip:
@martinthomson That doesn't answer the question "which formats are supported" and importantly can't tell you _which_ format among the supported ones is preferred.

martinthomson:
I don't know why you would need either enumeration or knowledge of browser preferences.  "Can you produce _X_?" is a question I can understand though.

yell0wd0g:
@martinthomson, both ways are necessary to specify what is supported, see the associated issue #59 where the rationale is presented.

martinthomson:
@miguelao that only says what you want, not _why_.

yell0wd0g:
I see two reasons. First is what @sole mentions in the bug: developers would like to offer users the list of supported codec/containers (this could be circumvented by the JS, say, trying a mega-list of codecs and containers against `isTypeSupported()`, but, really?). The second reason, and IMHO the most important, is to give hints to the JS as to which codecs might be supported by hardware acceleration, so they can, free willingly , prioiritise those (and there is currently no workaround for that).

yell0wd0g:
Ping

MediaRecorder.html
```diff
@@ -442,6 +443,46 @@
...
+             </a> [[!HTML5]]. The codecs parameter [[!RFC6381]], if necessary to
+             specify exactly how the resource is encoded, SHOULD be added
+             to <code>type</code>.
+             If <var>type</var> can be expressed in several equivalent ways
+             (e.g. with or without quotes and whitespace, or different
+             capitalization), the shortest form with the largest number of
+             lowercase characters MUST be used.
+
+             </a>.
+             <var>resultList</var> MAY be ordered by the UA so that MIME types
```
martinthomson, line 471: MAY is too wishy washy.  If the intent is to declare preferences, then say that: i.e.

> the <a>user agent</a> MUST place codecs that it prefers before other codes

MediaRecorder.html
martinthomson: I'm not seeing much reason to address this problem in this PR.

martinthomson (commented):
The biggest problem I see with this PR is the increased fingerprinting surface area.  This is a fairly significant increase, so you should be clear about the impact of the API.

MediaRecorder.html
foolip: How will you write tests and get interoperable implementations if the spec doesn't define the format of the returned strings?

fluffy:
I don't think this works. The problem is that without the codec parameters, this is not much help. If all it returns is "video/mp4",  - the apps  has no idea if I can get opus or not. If it returns the parameters,  then the list is nearly infinitely long. 

What I would propose is return three sets of mime types - the containers supports, the audio codecs supported, and the video codecs supports. And by mime type I mean something like "audio/opus" not any parameters. Then allow  the request for what to use to include parameters.

foolip:
Not all combinations of containers and codecs will be supported, so maybe return a map from container strings to array of codec strings? Not sure what the IDL for that looks like, I'm afraid.

yell0wd0g:
@martinthomson, @fluffy, I see value in this added method, let's try to reach some common ground :smile:
- the fingerprinting surface would not be increased: a motivated agent could always try a (large) number of mime types against the static function `MediaRecorder.isTypeSupported()` and retrieve the same information.  Do you think listing them as preferred by the platform could increase this risk?
- @fluffy, the idea is to enumerate all supported `MIMETypes`; the parameters should never be enumerated since they are codec-dependent and anyway an implementation detail.
  
  An example of this method for the Chrome implementation would retrieve the following list (reordered to match the platform particularities and preferences):
  
  ```
  video/webm
  video/webm;codecs=vp8
  video/webm;codecs=vp9
  video/webm;codecs=h264
  video/webm;codecs=opus
  audio/webm
  audio/webm;codecs=opus
  ```
  
  like @foolip says, the combinations have to be explicitly listed: in this particular case `video/webm; codecs=opus` is not a mistake but a specific [quirk](media/capture/video/win/sink_filter_win.h).

yell0wd0g:
@martinthomson please take a look at my comments right before this one (I misspelled the last letter of you id so prob. you didn't get any notification 😅 )

MediaRecorder.html
```diff
@@ -442,6 +443,46 @@
...
+            <li>Run the following steps in parallel:
+             <ol>
+             <li>Let <var>resultList</var> be an empty list.
+             <li>For each <a href="https://tools.ietf.org/html/rfc2046">MIME
+             type</a> <var>type</var> that is supported for encoding and/or
+             multiplexing (i.e. <code>isSuportedType(<var>type</var>)</code>
+             is true), add <var>type</var> to <var>resultList</var>.
+
+             <var>type</var> MUST be a <a def-id="mime-types">valid MIME type
+             </a> [[!HTML5]]. The codecs parameter [[!RFC6381]], if necessary to
```
ddorwin, line 462: The codecs parameter is almost always necessary. There are few MIME types that specify the codec without the codecs parameter. (One example is "audio/mpeg" for MP3.) There should be stronger language and requirements (not `SHOULD`) here. For example, EME [uses](https://w3c.github.io/encrypted-media/#get-supported-capabilities-for-audio-video-type) the following text:

> If _container_ normatively implies a specific set of codecs and codec constraints

MediaRecorder.html
```diff
@@ -442,6 +443,46 @@
...
+          <dd>
+            When this method is invoked, the User Agent must run the following
+            steps:
+            <ol>
+            <li>Let <var>p</var> be a promise.</li>
+            <li>Run the following steps in parallel:
+             <ol>
+             <li>Let <var>resultList</var> be an empty list.
+             <li>For each <a href="https://tools.ietf.org/html/rfc2046">MIME
+             type</a> <var>type</var> that is supported for encoding and/or
```
ddorwin, line 457: Is it expected that implementations will only support/expose a small number of encoding configurations (per codec)?

Codec strings (see my comment below) are sometimes (including in newer codecs) very exact. For example, `avc1.42E01E`. ~~Note that AVC1 requires the full six characters after `avc1.`, so returning `avc1` would not be spec-compliant.~~ Thus, this could be a very long list unless implementations limit what they expose.

In addition, the AVC1 spec seems to have multiple ways to encode the same information in the codec string. All this complexity could make it difficult for authors to use the information returned by such an API.

MediaRecorder.html
```diff
@@ -442,6 +443,46 @@
...
+             specify exactly how the resource is encoded, SHOULD be added
+             to <code>type</code>.
+             If <var>type</var> can be expressed in several equivalent ways
+             (e.g. with or without quotes and whitespace, or different
+             capitalization), the shortest form with the largest number of
+             lowercase characters MUST be used.
+
+             </a>.
+             <var>resultList</var> MAY be ordered by the UA so that MIME types
+             closer to the beginning are preferred, for example due to larger
```
ddorwin, line 472: This doesn't allow the application to distinguish which are actually encoded or otherwise preferred. For example, if there are three results, all, two, one, or none of them might be accelerated. Applications may end up just picking the first result.

Perhaps this should use the same mechanism as Media Capabilities ends up using. See the [discussion](https://github.com/WICG/media-capabilities/blob/master/explainer.md#decoding-capabilities) of smoothness and power efficiency in the explainer.

MediaRecorder.html
```diff
@@ -442,6 +443,46 @@
...
+            When this method is invoked, the User Agent must run the following
+            steps:
+            <ol>
+            <li>Let <var>p</var> be a promise.</li>
+            <li>Run the following steps in parallel:
+             <ol>
+             <li>Let <var>resultList</var> be an empty list.
+             <li>For each <a href="https://tools.ietf.org/html/rfc2046">MIME
+             type</a> <var>type</var> that is supported for encoding and/or
+             multiplexing (i.e. <code>isSuportedType(<var>type</var>)</code>
```
ddorwin, line 458: `isTypeSupported`

MediaRecorder.html
```diff
@@ -80,6 +80,7 @@
     void           resume ();
     void           requestData ();
     static boolean isTypeSupported (DOMString type);
```
ddorwin, line 82: See below regarding MIME types and codec strings. Unless implementations support encoding many profiles and levels, it will be difficult for applications to find exact matches. For example, `avc1.42E01E` vs. `avc1.4D401E`. For other such methods, the application has a stream and wants to know if the user agent supports it. In this case, the application must try to guess at the strings the user agent might support (or be strict in only support specific profiles and levels, which could limit client compatibility).

More generally, the spec should be more specific about checking `type`, including generally requiring `codecs`, checking audio vs. video types, and determining whether multiple codecs (e.g. audio and video, if applicable) are supported together. See, for example, how EME [handles MIME content type strings](https://w3c.github.io/encrypted-media/#get-supported-capabilities-for-audio-video-type).

Three of the four types in Example 7.1 are incorrect. Two only specify the container, and another uses `h264` rather than `avc1.xxxxxx`. Also, these are content types, not "codecs," which is the variable name.

MediaRecorder.html
ddorwin: Correction: `avc` is compliant according to https://tools.ietf.org/html/rfc6381#section-3.4. However, it is not very descriptive since AVC supports lots of profiles, most of which are unlikely to be supported. Also, this does not allow, for example, differentiating between baseline and high profile.

See item (2) in #93.

MediaRecorder.html
ddorwin: I filed #92, #93, and #94.

MediaRecorder.html
ddorwin: See #92, which proposes to be more specific in the `isTypeSupported()` algorithm. The text should be similar and/or reuse algorithm(s).

yell0wd0g:
Let's move the discussion to the issue https://github.com/WICG/media-capabilities/issues/28.

PR #66, opened 2016-08-03
Use https links for WG home pages

dontcallmedom:
For consistency with new pubrules https://lists.w3.org/Archives/Public/spec-prod/2016JulSep/0022.html


yell0wd0g:
this very much LGTM

Issue #79, opened 2016-10-07
Clarify ```ignoreMutedMedia```


stefhak:
`ignoreMutedMedia` talks about a muted _MediaStream_, however, a MediaStream can't be muted, so perhaps it should say _MediaStreamTrack_?

Further, the last sentence speccing the `start` methods says that muted tracks will be recorded as silence/blackness, but should not `ignoreMutedMedia` make a difference?

yell0wd0g:
@stefhak please take a look and close this bug if satisfied, cheers!

stefhak:
LGTM! Closing.

PR #80, opened 2016-10-31
Corrections re. |ignoreMutedMedia|

yell0wd0g:
Addresses #79, usage of |ignoreMutedMedia|:
- |muted| is an attribute of a `MediaStreamTrack`
- `start()` method should refer (and link) to the |ignoreMutedMedia|.

Couple of other tiny things, e.g.  instead of 
`<em title="must" class="rfc2119">must</em> `, `MUST` is enough.
Text is reflowed.


yell0wd0g:
Given that this is a tiny syntactical correction, I'm merging it.

PR #81, opened 2016-10-31
Cleanups, reflows and addressing some editorial issues.

yell0wd0g:


MediaRecorder.html
```diff
@@ -215,37 +231,40 @@
               <li>Set <code>state</code> to "recording", and run the following
               steps in parallel:
                 <ol>
+
                   <li>Once media becomes available from one or more of the
                   stream's tracks, start gathering the data into a Blob (see
                   [[!FILE-API]]), and queue a task, using the DOM manipulation
-                  task source, to fire a <code>start</code> event.</li>
+                  task source, to fire an event named <code>start</code>.</li>
```
foolip, line 238: Can you link to https://dom.spec.whatwg.org/#concept-event-fire?

MediaRecorder.html
```diff
@@ -254,11 +273,12 @@
                   following steps:
                     <ol>
                       <li>set <code>recording</code> to "inactive".</li>
-                      <li>Fire a <code>dataavailable</code> event containing
-                      <var>blob</var>.</li>
-                      <li>Fire a <code>stop</code> event.</li>
+                      <li>Fire an event named <code>dataavailable</code>
```
foolip, line 276: Based on https://dom.spec.whatwg.org/#firing-events-example I think this can be "Fire an event named `dataavailable` at _target_ using `BlobEvent` with its `blob` event intialized to _blob_" where _target_ and _blob_ are variables to be defined in previous steps.

Note that the `ErrorEvent` cases can also be rewritten on that form, although as long as it's unambiguous it doesn't matter if there's a "fire an error event" concept.

Issue #78, opened 2016-10-05
Incorrect use of mime type 


fluffy:
The draft calls things that are more than a mime type a mine type. This should be correct and it should reference rfc6381.  Right now were it claims to take a mime type, it can probably only really take a container mine type with possible parameters that specify the mine types for the codecs used by the container.

yell0wd0g:
Also touched by #83.

yell0wd0g:
@fluffy can you take a look at the wording in https://w3c.github.io/mediacapture-record/MediaRecorder.html#dom-mediarecorder-mimetype and https://w3c.github.io/mediacapture-record/MediaRecorder.html#dom-mediarecorderoptions-mimetype and see if they address your concerns? Thx

yell0wd0g:
ping @fluffy.

yell0wd0g:
reping @fluffy.  Closing this issue preventively, please reopen if not satisfied.

Issue #55, opened 2016-04-16
Can .start() throw a SecurityError?


jan-ivar:
The following text is quite troubling it if doesn't (emphasis mine):

> "If at any point the stream's isolation properties change so that MediaRecorder is no longer allowed access to it, the UA must stop recording, discard any data that it has gathered and throw a SecurityError DOMException followed by the stop event. If the UA is unable to start recording or at any point is unable to continue recording **_(i.e., for reasons other than a security violation)**_, it must throw an UnknownError DOMException, followed by a dataavailable event containing the Blob it has gathered, follwed by the stop event."

(Nevermind that it's impossible to "throw a SecurityError" for runtime errors, I think it means fire the `error` event.)

If `.start()` doesn't throw `SecurityError` then this is a hole big enough to drive an already isolated stream through (since a stream doesn't really "change" if it starts out, and stays, isolated).

jan-ivar:
Unfortunately, the algorithm for `.start` has no mention of throwing `SecurityError`.

jan-ivar:
There's two ways to solve this.

This came up because Firefox recently changed its behavior: `.start` used to throw `SecurityError`, but due to a code reorganization, it now fires the `error` event instead. Almost the same effect, just one fewer code-path to handle. I was told there wasn't any technical necessity behind the change, so we could go either way.

martinthomson:
In the interest of making errors available in the context in which they can be most readily acted upon, I think that firing `SecurityError` from `start()` makes the most sense to me.

jan-ivar:
@martinthomson good principle. I buy it.

yell0wd0g:
@jan-ivar can you take a look at #83 wording (or the status of the document after this PR lands) and see if sth/what needs to be done? Thx!

jan-ivar:
@miguelao This particular issue was already addressed by https://github.com/w3c/mediacapture-record/pull/61 so we should be good here.

Issue #54, opened 2016-04-14
Async algorithms .start(), .stop(). pause() et al. need major work


jan-ivar:
The fundamental design of async algorithms should be:
1. You do some stuff synchronously up front to make sure the API call makes sense, and throw if not.
2. You return.
3. You do async stuff like waiting, fetching, etc. [_in parallel_](https://html.spec.whatwg.org/#in-parallel).
4. Any further communication with your original caller happens from tasks you post to a task source.

Compare this to the `.start()` algorithm (and most of the others):
1. Queues a task immediately on the _DOM manipulation task source_ (effectively the main thread!).
2. It never returns.
3. It throws exceptions from a queued task, which never make it back to content.
4. It blocks the main thread with language like "wait until media becomes available".
5. It fires events like `stop` synchronously from code that seems asynchronous, causing JS concurrency.

I suspect all the algorithms that queue to _DOM manipulation task source_ and then wait should be rewritten this way. Some like `.resume()` may be fine as-is, though I'm curious why we seem to prefer changing state on the next event loop rather than synchronously?

yell0wd0g:
@jan-ivar, with #56 and the rewriting in #83, what's still needed for this issue to be addressed?

jan-ivar:
@miguelao These still need to be addressed for all methods other than _start_. See my review comments in https://github.com/w3c/mediacapture-record/pull/83 for details.

yell0wd0g:
Rewrote the algorithms so they read more like the following pseudocode:
- `the UA must run the following steps`:
- something is wrong upfront: `throw an` bla `DOMException, else`queue a task, using the DOM manipulation task source, that runs the following steps:`
  - do stuff
  - something goes wrong: `fire an event` named whatever, etc.

is this closer to what you meant?

yell0wd0g:
@jan-ivar could you please have another look at the current `HEAD`? We can clean it up, if needed, from there.

jan-ivar:
Done. lgtm. comments in https://github.com/w3c/mediacapture-record/issues/83

yell0wd0g:
Please take another look at the current status, thanks!

jan-ivar:
lgtm.

yell0wd0g:
Closing per https://github.com/w3c/mediacapture-record/issues/54#issuecomment-258252176

Issue #42, opened 2015-12-22
Consider merging ErrorEvent definition with Media Capture and Streams


yell0wd0g:
@beverloo insightfully commented in https://github.com/w3c/mediacapture-record/issues/31#issuecomment-166599507 

> https://html.spec.whatwg.org/multipage/webappapis.html#the-errorevent-interface
> http://w3c.github.io/mediacapture-main/#errorevent (duplicated [here](https://rawgit.com/w3c/mediacapture-record/master/MediaRecorder.html#errorevent) — could we introduce a dependency instead?)
> 
> This seems unfortunate. The media one is a subset of the HTML one, compatible due to the nature of any error in both the interface and the init dictionary.

yell0wd0g:
Also filed as https://github.com/w3c/mediacapture-main/issues/298 so both repos have it.

jan-ivar:
See https://github.com/w3c/mediacapture-main/issues/335

yell0wd0g:
Closing this one since w3c/mediacapture-main#298 and w3c/mediacapture-main#335 are closed and there seems to be consensus that [`ErrorEvent`](https://html.spec.whatwg.org/multipage/webappapis.html#the-errorevent-interface), with its `filename`, `lineno` and `colno` is not useful in the current context.

PR #83, opened 2016-11-01
Follow-up @foolip's suggestion in #81: 'firing' and other literatures.

yell0wd0g:
- Made a section for the explicit definition of what
 'Firing a dataavailable event' means and use it elsewhere.
- Linking 'fire' keyword to DOMSpec's concept-event-fire
- Cleaned up references to 'state' and the enumeration entries.
- Reflowed some sequential entries into lists.
- Removed MediaRecorder.css (unneeded)

General reflow caused by adding href's


yell0wd0g:
@foolip do you want to take a look?

MediaRecorder.html
```diff
@@ -193,9 +193,10 @@
           href="https://www.w3.org/TR/mediacapture-streams/#mediastreamtrack">
           <code>MediaStreamTrack</code></a> is muted. If this attribute is
           <code>false</code>, MediaRecorder will record silence (for audio) and
-          black frames (for video) when the input MediaStream is muted. When the
-          MediaRecorder is created, the UA MUST set this attribute to
-          <code>false</code>.</dd>
+          black frames (for video) when the correspondent MediaStreamTrack is
```
jan-ivar: corresponding

MediaRecorder.html
```diff
@@ -33,15 +32,15 @@
...
-    one time. By default a <a
+    instantiates a <code><a>MediaRecorder</a></code> object, calls
+    <a href="#dom-mediarecorder-start">start()</a> and then calls
+    <a href="#dom-mediarecorder-stop">stop()</a> or waits for the
+    MediaStreamTrack(s) to be ended. The contents of the recording will be made
+    available in the platform's default encoding via the <a
+    href="#dom-mediarecorder-ondataavailable">ondataavailable</a> event.
+    Functions are available to query the platform's available set of encodings,
+    and to select the desired ones if the author wishes. The application can
+    also choose how much data it wants to receive at one time. By default a <a
```
jan-ivar, line 43: I don't see any changes here. Not sure about this spec, but in other specs I know we sometimes try to leave formatting alone, or at least do it in separate PRs since it makes git blame harder to follow. YMMV

MediaRecorder.html
```diff
@@ -67,26 +66,26 @@
...
+                          attribute boolean        ignoreMutedMedia;
+          readonly        attribute unsigned long  videoBitsPerSecond;
+          readonly        attribute unsigned long  audioBitsPerSecond;
+          void           start (optional long timeslice);
+          void           stop ();
+          void           pause ();
+          void           resume ();
+          void           requestData ();
+          static boolean isTypeSupported (DOMString type);
+      };</pre>
```
jan-ivar: Same here. Is there any appreciable difference from this change?

MediaRecorder.html
```diff
@@ -222,61 +223,83 @@
...
+                <ol>
                   <li>Once media becomes available from one or more of the
-                  stream's tracks, start gathering the data into a Blob (see
-                  [[!FILE-API]]), and queue a task, using the DOM manipulation
-                  task source, to fire an event named <code>start</code>.</li>
+                  stream's tracks, let <var>blob</var> be the Blob
+                  ([[!FILE-API]]) of collected data, start gathering the data
+                  into <var>blob</var> and queue a task, using the DOM
+                  manipulation task source, to <a
+                  href="https://dom.spec.whatwg.org/#concept-event-fire">fire
```
foolip, line 245: This and other places should also ideally say what the event target is. "the MediaRecorder instance" maybe if you don't have it in a variable. Or you could say "Let _foo_ be the context object" at the API entry points and use _foo_ in places like this.

MediaRecorder.html
```diff
@@ -222,61 +223,83 @@
...
                   it has gathered, and queue a task, using the DOM manipulation
-                  task source, that <a>fires an error event</a> with the name
-                  <code>SecurityError</code>, immediately followed by a <code>
-                  stop</code> event. </li>
+                  task source, that runs the following steps:
+                  <ol>
+                    <li>Set <a>state</a> to
+                    <a href="#dom-recordingstate-inactive">inactive</a>.</li>
+                    <li><a href="#dfn-firing-an-error-event-named-e">Fire an
+                    error event</a> with the name <code>SecurityError</code>
```
foolip: https://w3c.github.io/mediacapture-record/MediaRecorder.html#dfn-firing-an-error-event-named-e takes as its argument an Error _error_, so this should probably be 'Fire an error event with a `SecurityError` DOMException.' And see https://github.com/w3c/mediacapture-record/issues/31#issuecomment-166599507 and linked issues for the problem of `ErrorEvent` conflicting with HTML.

MediaRecorder.html
```diff
@@ -469,6 +498,22 @@
         </dl>
       </section>
     </div>
+    <section>
+      <h2>Data handling</h2>
+        <p><dfn data-dfn-type="dfn" data-lt="fires a dataavailable event">
```
foolip: The data-lt="" here seems wrong. Typo?

MediaRecorder.html
```diff
@@ -222,61 +223,83 @@
...
+                  <ol>
+                    <li>Set <a>state</a> to
+                    <a href="#dom-recordingstate-inactive">inactive</a>.</li>
+                    <li><a href="#dfn-firing-an-error-event-named-e">Fire an
+                    error event</a> with the name <code>UnknownError</code>
+                    </li>
+                    <li><a href="#dfn-firing-a-dataavailable-event">Fire a
+                    dataavailable event</a> with <var>blob</var>.</li>
+                    <li><a href="https://dom.spec.whatwg.org/#concept-event-fire">
+                    Fire</a> an event named <code>stop</code>.</li>
```
foolip: Here and elsewhere I'd suggest making all of "fire an event" the linked text. It doesn't really matter, but for consistency and if you ever convert this to Bikeshed that's what you'll use to get the linked right automatically.

MediaRecorder.html
```diff
@@ -222,61 +223,83 @@
...
+                    <a href="#dom-recordingstate-inactive">inactive</a>.</li>
+                    <li><a href="#dfn-firing-an-error-event-named-e">Fire an
+                    error event</a> with the name <code>SecurityError</code>
+                    </li>
+                    <li><a href="https://dom.spec.whatwg.org/#concept-event-fire">
+                    Fire</a> an event named <code>stop</code>.</li>
+                  </ol>
+                  </li>
 
                   <li>If the UA at any point is unable to continue gathering
```
foolip, line 267: Oddly enough there are no existing cases with which to be consistent, so I guess you get to pick the style for the whole file, but now that this is no longer a single paragraph, I'd probably put the first bit in `<p>` on a new line.

MediaRecorder.html
```diff
@@ -370,18 +397,19 @@
...
             <ol>
-              <li>If <code>state</code> is "inactive" throw an
-              <code>InvalidStateError</code> DOMException and terminate these
-              steps. Otherwise:</li>
-              <li>Set <code>state</code> to "recording".</li>
-              <li>Resume (or continue) gathering data into its current
-              Blob.</li>
-              <li>Fire an event named <code>resume</code>.</li>
+              <li>If <a>state</a> is <a href="#dom-recordingstate-inactive">
+              inactive</a> throw an <code>InvalidStateError</code> DOMException
```
foolip, line 420: You only changed the formatting, but this doesn't make sense. Inside a task, what does it mean to throw an exception? In Blink implementation terms, you'd have no `ExceptionState` available.

MediaRecorder.html
```diff
@@ -469,6 +498,22 @@
...
       </section>
     </div>
+    <section>
+      <h2>Data handling</h2>
+        <p><dfn data-dfn-type="dfn" data-lt="fires a dataavailable event">
+        Firing a dataavailable event </dfn> with a <a
+        href="https://www.w3.org/TR/FileAPI/#blob"><code>Blob</code></a>
+        <var>blob</var> means <a
+        href="https://dom.spec.whatwg.org/#concept-event-fire">firing</a>
+        an event named <code>dataavailable</code> at the current
```
foolip: I'd make the target an input to this algorithm for consistency with DOM's "fire an event" so that the call sites look pretty much the same.

MediaRecorder.html
```diff
@@ -469,6 +498,22 @@
         </dl>
       </section>
     </div>
+    <section>
+      <h2>Data handling</h2>
+        <p><dfn data-dfn-type="dfn" data-lt="fires a dataavailable event">
+        Firing a dataavailable event </dfn> with a <a
```
foolip: I think this should be "fire a blob event", as specializations of "fire an event" typically use the event interface name and not the event type. So the call sites would say "fire a blob event named dataavailable at _target_ with its `blob` attribute initialized to _blob_" or some such.

MediaRecorder.html
```diff
@@ -495,7 +540,14 @@
...
             platform-specific default format. The container format, whether
             passed in to the constructor or defaulted, will be used as the value
-            of the <code>mimeType</code> attribute.</dd>
+            of the <code>mimeType</code> attribute.
+
+            <div class="note"> <code><a>mimeType</a></code> specifies the media
+            type and container format for the recording via a type/subtype
+            combination, with the codecs and/or profiles parameters [[RFC6381]]
+            specified where ambiguity might arise. Individual codecs might have
+            further optional specific parameters. </div></dd>
```
foolip, line 572: Might want to run this through a validator to see if there's not one `</dd>` too many now.

MediaRecorder.html
```diff
@@ -565,10 +619,11 @@
     <h2>Blob Event</h2>
     <div>
       <pre class="idl">
-      [ Constructor (DOMString type, BlobEventInit eventInitDict)]
-interface BlobEvent : Event {
-    readonly        attribute Blob data;
-};</pre>
+        [ Constructor (DOMString type, BlobEventInit eventInitDict)]
```
foolip: If you're changing this anyway, I'd remove the whitespace after `[` and all the extra whitespace before `attribute`.

MediaRecorder.html
```diff
@@ -222,61 +223,83 @@
...
                   <li>Once media becomes available from one or more of the
-                  stream's tracks, start gathering the data into a Blob (see
-                  [[!FILE-API]]), and queue a task, using the DOM manipulation
-                  task source, to fire an event named <code>start</code>.</li>
+                  stream's tracks, let <var>blob</var> be the Blob
+                  ([[!FILE-API]]) of collected data, start gathering the data
+                  into <var>blob</var> and queue a task, using the DOM
+                  manipulation task source, to <a
+                  href="https://dom.spec.whatwg.org/#concept-event-fire">fire
+                  </a> an event named <code>start</code>.</li>
```
jan-ivar: Unsure of the value of using let just to use it in the same sentence.

MediaRecorder.html
```diff
@@ -648,17 +703,24 @@
...
-      error, the UA MUST fire an event named <code>dataavailable</code>,
-      containing any data that it has gathered, and then a <code>stop</code>
-      event. The UA may set platform-specific limits, such as those for the
-      minimum and maximum Blob size that it will support, or the number of
-      Tracks it will record at once. It must signal a fatal error if these
-      limits are exceeded.</p>
+
+      <p>The UA MUST throw a <a
+      href="https://heycam.github.io/webidl/#idl-DOMException">DOMException</a>
+      ([[!DOM4]]) when the error can be detected at the time that the call is
```
foolip: Not sure why this says DOM4, the link is right in that WebIDL defines DOMException now.

MediaRecorder.html
jan-ivar: I prefer the old wording here actually. We also define _blob_ in step 4, so I think we should avoid having the same named variable twice.

MediaRecorder.html
```diff
@@ -222,61 +223,83 @@
...
-                  following steps:
-                    <ol>
-                      <li>set <code>recording</code> to "inactive".</li>
-                      <li>Fire an event named <code>dataavailable</code>
-                      containing <var>blob</var>.</li>
-                      <li>Fire an event named <code>stop</code>.</li>
-                    </ol>
+                  <li>If all recorded tracks become <a
+                  href="http://www.w3.org/TR/mediacapture-streams/#track-ended">
+                  ended</a> [[!GETUSERMEDIA]], then stop gathering data and
```
jan-ivar: important comma missing (we don't want to stop queuing a task).

MediaRecorder.html
```diff
@@ -330,14 +353,16 @@
...
-              <code>InvalidStateError</code> DOMException and terminate these
-              steps. Otherwise:</li>
-              <li>Set <code>state</code> to <code>inactive</code> and stop
-              gathering data.</li>
-              <li>Fire an event named <code>dataavailable</code> containing the
-              Blob of data that has been gathered.</li>
-              <li>Fire an event named <code>stop</code>.</li>
+              <li>If <a>state</a> is <a href="#dom-recordingstate-inactive">
+              inactive</a>, throw an <code>InvalidStateError</code> DOMException
+              and terminate these steps. Otherwise:</li>
```
jan-ivar: (Thanks for addressing _stop_) We should check the state synchronously and throw error synchronously, like we do on _start_, as throwing an error in a queued task does no good.

MediaRecorder.html
```diff
@@ -330,14 +353,16 @@
...
+              <li>Set <a>state</a> to <a href="#dom-recordingstate-inactive">
+              inactive</a> and stop gathering data.</li>
```
jan-ivar, line 372: We should also set the state synchronously. Both Firefox and Chrome do that.

MediaRecorder.html
```diff
@@ -330,14 +353,16 @@
...
+              <li>Let <var>blob</var> be the Blob of collected data so far, then
```
jan-ivar: To me, "so far" sounds like we'll get a blob with everything from the start, which is not the case. Maybe something like "the newest Blob of collected data", or something?

jan-ivar: I've always wondered, in practice, don't we need an "in parallel" step here, like in start, since gathering happens off main thread? I wonder the same thing about `requestData`.

MediaRecorder.html
```diff
@@ -330,14 +353,16 @@
...
+              <a href="#dfn-firing-a-dataavailable-event">fire a dataavailable
+              event</a> with <var>blob</var>.</li>
+              <li><a href="https://dom.spec.whatwg.org/#concept-event-fire">Fire
+              </a> an event named <code>stop</code>.</li>
```
jan-ivar: These last two steps are the only ones I think need to remain in the queued task.

MediaRecorder.html
```diff
@@ -353,13 +378,15 @@
...
-              <li>If <code>state</code> is "inactive" throw an
-              <code>InvalidStateError</code> DOMException and terminate these
-              steps. Otherwise:</li>
-              <li>Set <code>state</code> to "paused".</li>
-              <li>Stop gathering data into its current Blob (but keep the Blob
+              <li>If <a>state</a> is <a href="#dom-recordingstate-inactive">
+              inactive</a> throw an <code>InvalidStateError</code> DOMException
+              and terminate these steps. Otherwise:</li>
+              <li>Set <a>state</a> to <a href="#dom-recordingstate-paused">
+              paused</a>.</li>
```
jan-ivar, line 400: Same thing here, these two steps need to happen synchronously (throwing in a queued task is a no-op).

MediaRecorder.html
```diff
@@ -353,13 +378,15 @@
...
+              <li>Stop gathering data into <var>blob</var> (but keep it
               available so that recording can be resumed in the future).</li>
-              <li>Fire an event named <code>pause</code>.</li>
+              <li><a href="https://dom.spec.whatwg.org/#concept-event-fire">Fire
+              </a> an event named <code>pause</code>.</li>
```
jan-ivar: This is the only step that should remain in the queued task imho.

MediaRecorder.html
```diff
@@ -370,18 +397,19 @@
...
-              steps. Otherwise:</li>
-              <li>Set <code>state</code> to "recording".</li>
-              <li>Resume (or continue) gathering data into its current
-              Blob.</li>
-              <li>Fire an event named <code>resume</code>.</li>
+              <li>If <a>state</a> is <a href="#dom-recordingstate-inactive">
+              inactive</a> throw an <code>InvalidStateError</code> DOMException
+              and terminate these steps. Otherwise:</li>
+              <li>Set <a>state</a> to <a href="#dom-recordingstate-recording">
+              recording</a>.</li>
```
jan-ivar, line 425: Same thing with _resume_.

MediaRecorder.html
```diff
@@ -397,12 +425,13 @@
             task, using the DOM manipulation task source, that runs the
             following steps:
             <ol>
-              <li>If <code>state</code> is not "recording" throw an
-              <code>InvalidStateError</code> DOMException and terminate these
-              steps. Otherwise:</li>
```
jan-ivar, line 402: Same thing with `requestData`.

MediaRecorder.html
```diff
@@ -469,6 +498,22 @@
...
+        <p><dfn data-dfn-type="dfn" data-lt="fires a dataavailable event">
+        Firing a dataavailable event </dfn> with a <a
+        href="https://www.w3.org/TR/FileAPI/#blob"><code>Blob</code></a>
+        <var>blob</var> means <a
+        href="https://dom.spec.whatwg.org/#concept-event-fire">firing</a>
+        an event named <code>dataavailable</code> at the current
+        <code>MediaRecorder</code> object using a <a>BlobEvent</a> with its
+        <code>data</code> initialized to <var>blob</var>.
+        Usually <var>blob</var> will be the data gathered by the UA after the
+        last transition to <a href="#dom-recordingstate-recording">recording
```
jan-ivar, line 534: I don't think this last sentence holds since state doesn't change when _timeslice_ is up, or `requestData` is called. Maybe leave out that sentence?

MediaRecorder.html
```diff
@@ -648,17 +703,24 @@
...
+      <var>blob</var> be the Blob of collected data so far; after raising the
+      error, the UA MUST <a href="#dfn-firing-a-dataavailable-event">fire a
+      dataavailable event</a> with <var>blob</var>;
+      the UA MUST then
+      <a href="https://dom.spec.whatwg.org/#concept-event-fire">fire</a> a
+      <code>stop</code> event.
+      The UA may set platform-specific limits, such as those for the minimum and
+      maximum Blob size that it will support, or the number of Tracks it will
+      record at once.
+      It must signal a fatal error if these limits are exceeded.</p>
```
jan-ivar, line 746: Again I detect no change. Not a fan of reformatting for its own sake. Maybe once the individual methods have been fixed we can get rid of this section? A "general principles" section is an odd place for normative language IMHO.

MediaRecorder.html
```diff
@@ -670,10 +732,11 @@
...
       <code><a data-link-for= "ErrorEvent">error</a></code> attribute set to
-      <var>error</var>, must be created and <a href="https://dom.spec.whatwg.org
-      /#concept-event-dispatch">dispatched</a> at the given target. If no
-      <code>Error</code> object is specified, the <code><a data-link-for=
-      "ErrorEvent">error</a></code> attribute defaults to null.</p>
+      <var>error</var>, must be created and
+      <a href="https://dom.spec.whatwg.org/#concept-event-dispatch">dispatched
+      </a> at the given target. If no <code>Error</code> object is specified,
+      the <code><a data-link-for="ErrorEvent">error</a></code> attribute
+      defaults to null.</p>
```
jan-ivar: No change detected.

jan-ivar (commented):
Thanks for doing this. The problems of https://github.com/w3c/mediacapture-record/issues/54 remain however. See detailed comments.

MediaRecorder.html
yell0wd0g: Done

MediaRecorder.html
yell0wd0g: No semantic changes - the change to ReSpect justified all the idl's 
to the left and, since I'm text editing, I just correct them as I see 
them, knowing that they are easy to spot.

MediaRecorder.html
yell0wd0g: In the LHS some href links were broken due to line breaks -
this was likely my mistake in using the text editor's paragraph 
reallign. 
The changes include:
- s/instatiates/instantiates in the first line
- s/start/stop/ in the second href.
  The rest is probably just text reflow - it's unfortunate, but very
  common due to the long href links and the 80 character limit;
  some other Specs do not respect the 80 column limit, which
  causes other troubles during review: GitHub tends to show 
  those lines with horizontal scroll bars.

MediaRecorder.html
yell0wd0g: Done: 
`<a href="https://dom.spec.whatwg.org/#concept-event-fire">Fire an event</a>`

MediaRecorder.html

MediaRecorder.html
yell0wd0g: I just copied it from the similar 
`<dfn data-dfn-type="dfn" data-lt="fires an error event">`
which I think I've seen around the image capture spec as well.

MediaRecorder.html
yell0wd0g: </dd> is closing l.536 of the RHS (collapsed lines).

However, I'd like to validate the document, will address
 in issue #84 (clearer PR merging/history).

MediaRecorder.html
yell0wd0g: Why not? Done here and in the other idls.

MediaRecorder.html
yell0wd0g: Oh, seems to be an alias (http://www.specref.org/?q=DOM4#dom), changed to `[[DOM]]`

MediaRecorder.html
yell0wd0g: Rewritten these wrong method descriptions (`stop`, `pause`, `resume` and `requestData`) so that
they follow the pattern:
`the UA must run the following steps` --> if in wrong state then `throw DOMException` --> otherwise `the UA MUST queue a task, using the DOM manipulation task source, that runs the following steps:` --> etc.

The new wording also follows the Blink implementation.

MediaRecorder.html
yell0wd0g: I'll have a single `let <var>blob</var> be the Blob of collected data` and will
leave it here so later paragraphs can refer to it; leaving it very similar to
how it was before.

MediaRecorder.html
yell0wd0g: I should probably do a single PR with html cleanup and homogeneisation: #85.

MediaRecorder.html
yell0wd0g: Rewritten to the best of my knowledge

MediaRecorder.html

MediaRecorder.html
yell0wd0g: Done by means of specifying 
`... let <var>target</var> be the MediaRecorder context object`
where appropriate

MediaRecorder.html
yell0wd0g: I think I should address this wholesale change  in another
PR because -- I understand that using the wording 
`Fire an error event with a Bla DOMException` would bypass the local `ErrorEvent`.  

FWIW, #31 is still open.

Edit: Removed some nonsense from my previous reply.
I've been going over the options and see, summing up, that:
- `Media Capture and Streams` doesn't use Error Events of any kind beyond `OverConstrainedError` and that's not interesting for MediaRecorder.
- HTML's `ErrorEvent` does not apply to the MR case as discussed in #42 and  w3c/mediacapture-main#335. 

Options then:
- use an [`Event`](https://www.w3.org/TR/dom/#event) with some indication of it being an error (`type`?).
- use/subtype a [`CustomEvent`](https://www.w3.org/TR/dom/#customevent) with the error indication in the `detail` field .
- use a `use a DOMException` as an .. `Event` - as you seem to suggest, @foolip? What would be the wording of these, just remove the `ErrorEvent` section completely and write '`Fire an event` named `error` with a SecurityError DOMException' ...?

yell0wd0g (commented):
Most (all?) comments addressed in PS2, ptal :-)

MediaRecorder.html
foolip: Thanks, that makes much more sense :)

MediaRecorder.html
foolip: By 'Fire an error event with a SecurityError DOMException.' I didn't mean something like `target.dispatchEvent(new DOMException('', 'SecurityError'))`, I was just reacting to the fact that `SecurityError` is a string, not an Error or DOMExceptions.

foolip: Since this already didn't make any sense, handling it in a separate PR is probably best.

MediaRecorder.html
```diff
@@ -220,63 +222,87 @@
             <code>start()</code> method is invoked, the UA MUST run the
             following steps:
             <ol>
-              <li>Let <var>timeslice</var> be the method's first argument, if
-              provided, or <code>undefined</code>.</li>
-              <li>If <code>state</code> is not <code>inactive</code>, throw an
-              <code>InvalidStateError</code> DOMException and abort these
-              steps.</li>
+              <li>Let <var>target</var> be the MediaRecorder context object.
+              Let <var>timeslice</var> be the method's first argument, if
```
foolip: A new `<li>` for the new let?

MediaRecorder.html
```diff
@@ -220,63 +222,87 @@
...
+                <ol>
                   <li>Once media becomes available from one or more of the
-                  stream's tracks, start gathering the data into a Blob (see
-                  [[!FILE-API]]), and queue a task, using the DOM manipulation
-                  task source, to fire an event named <code>start</code>.</li>
+                  stream's tracks, start gathering the data into a Blob
+                  ([[!FILE-API]]) <var>blob</var> and queue a task, using the
+                  DOM manipulation task source, to <a
+                  href="https://dom.spec.whatwg.org/#concept-event-fire">fire
+                  an event</a> named <code>start</code>.</li>
```
foolip: This and other cases elsewhere still need "at _target_". I'm just imitating https://dom.spec.whatwg.org/#firing-events-example FWIW.

MediaRecorder.html
foolip: Although, a simple way to improve it in the interim is to make https://w3c.github.io/mediacapture-record/MediaRecorder.html#dfn-firing-an-error-event-named-e take an "exception name _error_" instead of an "`Error` _error_". See https://html.spec.whatwg.org/multipage/embedded-content.html#reject-pending-play-promises for a similar setup.

MediaRecorder.html
```diff
@@ -469,12 +510,27 @@
         </dl>
       </section>
     </div>
+    <section>
+      <h2>Data handling</h2>
+        <p><dfn data-dfn-type="dfn"
+        data-lt="fire a blob event">Firing a blob event</dfn> with a
```
foolip: Optional nit: this is consistent with "Firing an error event", but if you think that'll go away "To fire a blob event named _e_ at _target_ with a `Blob` _blob_, fire an event..." would be more pleasing.

MediaRecorder.html
```diff
@@ -469,12 +510,27 @@
...
+        <a href="https://www.w3.org/TR/FileAPI/#blob"><code>Blob</code></a>
+        <var>blob</var> means
+        <a href="https://dom.spec.whatwg.org/#concept-event-fire">firing
+        an event</a> at <var>target</var> using a <a>BlobEvent</a> with its
+        <code>data</code> initialized to <var>blob</var>.
```
foolip: "data attribute initialized"

MediaRecorder.html
```diff
@@ -469,12 +510,27 @@
...
+        Usually <var>blob</var> will be the data gathered by the UA after the
```
foolip, line 533: Put this comment in a note?

MediaRecorder.html
foolip: Ack!

MediaRecorder.html
```diff
@@ -648,37 +714,47 @@
...
-      <a><code>ErrorEvent</code></a> MUST be fired. If recording has been
-      started and not yet stopped when the error occurs, then after raising the
-      error, the UA MUST fire an event named <code>dataavailable</code>,
-      containing any data that it has gathered, and then a <code>stop</code>
-      event. The UA may set platform-specific limits, such as those for the
-      minimum and maximum Blob size that it will support, or the number of
-      Tracks it will record at once. It must signal a fatal error if these
-      limits are exceeded.</p>
+
+      <p>The UA MUST throw a <a
```
foolip, line 730: Nit for future PR: This isn't new, but this seems like it should all be non-normative to me. There are already steps requiring exceptions to be thrown and events to be fired, and if this section should happen to disagree those steps are presumably the ones that are right.

MediaRecorder.html
```diff
@@ -648,37 +714,47 @@
...
     </section>
     <section>
       <h3>ErrorEvent</h3>
       <p>The <a>ErrorEvent</a> interface is defined for cases when an event is raised
-      that could have been caused by an error.</p> <p><dfn data-dfn-type="dfn"
-      data-lt="fires an error event">Firing an error event named
+      that could have been caused by an error.</p>
+      <p><dfn data-dfn-type="dfn"
+      data-lt="firing an error event">Firing an error event named
       <var>e</var></dfn> with an <code>Error</code> <var>error</var> means that
```
foolip: I'd move the `<dfn>` to earlier to match how you've done blob events.

MediaRecorder.html
foolip: I use `git diff --word-diff -w` when in doubt, it's pretty good.

foolip (approved):
LGTM % nits

MediaRecorder.html
yell0wd0g: Done.

MediaRecorder.html
yell0wd0g: Corrected, now all read `Fire an` {`event/error event/blob event`} `at target`.

MediaRecorder.html

MediaRecorder.html
yell0wd0g: Changed both to `to fire`

MediaRecorder.html
yell0wd0g: Temporarily changed the wording in `#dfn-firing-an-error-event-named-e` to read `To fire an error event named e with a DOMException named error means...`, and the `interface ErrorEvent` now has a `DOMException`. Since neither FF nor Chrome implements this `interface`, I think it doesn't matter to have it

MediaRecorder.html
yell0wd0g: Done

MediaRecorder.html
yell0wd0g: Why wait? Made the `General principles` Section non-normative now since anyway is just an overview of how errors are handled (i.e. as soon as possible).

MediaRecorder.html

yell0wd0g (commented):
Done all those below

yell0wd0g:
Since this review is pretty large and GitHub code review is quite cumbersome in the way the comments line up with the (re)actions to those comments, I'm going to land this PR, and we'll keep working on the linked issues: #54, #55, #78.

MediaRecorder.html
```diff
@@ -327,17 +361,22 @@
...
-              gathering data.</li>
-              <li>Fire an event named <code>dataavailable</code> containing the
-              Blob of data that has been gathered.</li>
-              <li>Fire an event named <code>stop</code>.</li>
+              <li>If <a>state</a> is <a href="#dom-recordingstate-inactive">
+              inactive</a>, throw an <code>InvalidStateError</code> DOMException
+              and terminate these steps. Otherwise the UA MUST queue a task,
+              using the DOM manipulation task source, that runs the following
+              steps:</li>
+              <li>Set <a>state</a> to <a href="#dom-recordingstate-inactive">
```
jan-ivar, line 371: Nit: maybe s/following/remaining/ ? Otherwise I'm used to seeing queued steps indented one level and renumbered. Maybe just be me. Or maybe hoist first step into initial paragraph, dunno.

MediaRecorder.html
```diff
@@ -393,16 +439,21 @@
...
-              <code>InvalidStateError</code> DOMException and terminate these
-              steps. Otherwise:</li>
-              <li>Fire an event named <code>dataavailable</code> containing the
-              current Blob of saved data. (Note that this Blob will be empty if
-              no data has been gathered yet.)</li>
+              <li>If <a>state</a> is not <a href="#dom-recordingstate-recording">
+              recording</a> throw an <code>InvalidStateError</code> DOMException
+              and terminate these steps. Otherwise the UA MUST queue a task,
+              using the DOM manipulation task source, that runs the following
+              steps::</li>
```
jan-ivar, line 449: double-colon

MediaRecorder.html
jan-ivar: 👍 except it needs to be reworded then to not sound normative, e.g. s/MUST/will always/ or similar, otherwise it's contradictory.

jan-ivar (commented):
Lgtm. Just a few nits, and it still seems a bit unclear to me how much data is gathered "so far".

Issue #84, opened 2016-11-01
Doc's HTML fails to pass W3C validation


yell0wd0g:
See here: https://validator.w3.org/nu/?doc=https%3A%2F%2Fw3c.github.io%2Fmediacapture-record%2FMediaRecorder.html

yell0wd0g:
Validation is happy -- except two warnings that seem irreconcilable.

PR #91, opened 2016-11-03
Followup #83  jan ivar comments and html validation

yell0wd0g:
This touches #84 and #54.


yell0wd0g:
Since this is a low-risk PR I'll go ahead and merge it.

Issue #73, opened 2016-10-04
Support for timecodes


sole:
An "Anonymous Developer" asked me for a way to insert [timecodes](https://en.wikipedia.org/wiki/Timecode) in the captured streams. They also asked for the counterpart: being able to read timecodes from streams (be it getUserMedia originated streams, or new streams created with MediaCapture, etc).

This would allow for video editing software to be built right in the browser, as timecodes allow for better accuracy for frame seeking than `getCurrentTime()`.

Apologies in advance, but I'm not quite sure if this should be filed here or somewhere else where media streams are defined.

yell0wd0g:
Note that this information is part of the container per-frame headers that are inside the `data` field. This information is, however, hard to get to -- essentially it means demultiplexing quite some parts of the output.

That said, we could have a [`DOMTimeStamp`](https://www.w3.org/TR/WebIDL/#common-DOMTimeStamp) field in the [`BlobEvent`](file:///Users/mcasas/p/github/mediacapture-record/MediaRecorder.html#blob-event):

```
interface BlobEvent : Event {
  //...
  readonly attribute DOMTimeStamp timeStamp;
};
```

The `timeStamp` would not necessarily refer to the inner `data` but to the `Event` itself (i.e. if `data` was a chunk encoded+multiplexed video frames and audio buffers, `timeStamp` would refer to the first one). WDYT, @foolip, @fluffy ?

foolip:
Something like that might work, but note that Event already has a timeStamp attribute, so it'd have to be some other name. I also guess it'd be a point on a timeline and not really a date-and-time?

yell0wd0g:
Then maybe just reuse the original terminology:

```
interface BlobEvent : Event {
  //...
  readonly attribute DOMTimeStamp timecode;
};
```

Also, note that `DOMTimestamp`is defined as:

> The DOMTimeStamp type is used for representing a number of milliseconds, either as an absolute time (relative to some epoch) or as a relative amount of time. Specifications that use this type will need to define how the number of milliseconds is to be interpreted.

which I interpret essentially as a large integer detailing milliseconds and not necessarily a time from some epoch.

foolip:
Uh, right, it's on the same timeline as `performance.now()`, not a date-time.

yell0wd0g:
@sole step 1 done.  Now for the implementations.

PR #95, opened 2016-11-28
First draft of adding |timecode| to BlobEvent

yell0wd0g:
Addresses #73.

### DIFF
--- a/MediaRecorder.html
+++ b/MediaRecorder.html
@@ -654,7 +654,7 @@
         [Constructor (DOMString type, BlobEventInit eventInitDict)]
         interface BlobEvent : Event {
           [SameObject] readonly attribute Blob data;
-          [SameObject] readonly attribute DOMHighResTimeStamp timecode;
+          readonly attribute DOMHighResTimeStamp timecode;
         };
       </pre>
       <section>
@@ -711,13 +711,20 @@
           <dd>The encoded blob. Its
           <a href="https://www.w3.org/TR/FileAPI/#dfn-type">
           type</a> attribute correlates to the selected
-          <a href="#dom-mediarecorder-mimeType">mimeType</a>.</dd>
+          <a href="#dom-mediarecorder-mimetype">mimeType</a>.</dd>
 
           <dt><dfn><code>timecode</code></dfn> of type <span class "idlAttrType">
           <a href="https://www.w3.org/TR/hr-time-2/#dom-domhighrestimestamp">
           DOMHighResTimeStamp</a></span>, readonly</dt>
-          <dd><code>Timecode</code> associated with the Blob data, relative to
-          the first produced <code>blob</code>.</dd>
+          <dd>Timecode associated with <var>data</var>.
+          The timestamp MUST be relative to the first produced
+          <a>BlobEvents</a>'s <a>timecode</a> (note that this value does not
+          need to be zero).
+          If <var>data</var> is composed of several chunks that could have
+          independent timecodes (e.g. several encoded video frames), then
+          <code>timecode</code> MUST refer to the earliest individually
+          addressable chunk.
+          </dd>
         </dl>
       </section>
     </div>
@@ -726,6 +733,7 @@
       <pre class="idl">
         dictionary BlobEventInit {
           required Blob data;
+          DOMHighResTimeStamp timecode;
         };
       </pre>
       <section>
@@ -735,6 +743,12 @@
           <dt><dfn><code>data</code></dfn> of type <span class=
           "idlMemberType"><a>Blob</a></span>, required</dt>
           <dd>A Blob object containing the data to deliver via this event.</dd>
+
+          <dt><dfn><code>timecode</code></dfn> of type <span class "idlMemberType">
+          <a href="https://www.w3.org/TR/hr-time-2/#dom-domhighrestimestamp">
+          DOMHighResTimeStamp</a></span>, optional</dt>
+          <dd>The timecode to be used in initializing <a>BlobEvent</a>.</dd>
+
         </dl>
       </section>
     </div>
--- a/MediaRecorder.html
+++ b/MediaRecorder.html
@@ -653,8 +653,8 @@
       <pre class="idl">
         [Constructor (DOMString type, BlobEventInit eventInitDict)]
         interface BlobEvent : Event {
-          readonly attribute Blob data;
-          readonly attribute DOMHighResTimeStamp timecode;
+          [SameObject] readonly attribute Blob data;
+          [SameObject] readonly attribute DOMHighResTimeStamp timecode;
         };
       </pre>
       <section>
--- a/MediaRecorder.html
+++ b/MediaRecorder.html
@@ -713,7 +713,7 @@
           type</a> attribute correlates to the selected
           <a href="#dom-mediarecorder-mimetype">mimeType</a>.</dd>
 
-          <dt><dfn><code>timecode</code></dfn> of type <span class "idlAttrType">
+          <dt><dfn><code>timecode</code></dfn> of type <span class="idlAttrType">
           <a href="https://www.w3.org/TR/hr-time-2/#dom-domhighrestimestamp">
           DOMHighResTimeStamp</a></span>, readonly</dt>
           <dd>The difference between the timestamp of the first chunk in
@@ -741,7 +741,7 @@
           "idlMemberType"><a>Blob</a></span>, required</dt>
           <dd>A Blob object containing the data to deliver via this event.</dd>
 
-          <dt><dfn><code>timecode</code></dfn> of type <span class "idlMemberType">
+          <dt><dfn><code>timecode</code></dfn> of type <span class="idlMemberType">
           <a href="https://www.w3.org/TR/hr-time-2/#dom-domhighrestimestamp">
           DOMHighResTimeStamp</a></span>, optional</dt>
           <dd>The timecode to be used in initializing <a>BlobEvent</a>.</dd>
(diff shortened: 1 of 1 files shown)


yell0wd0g:
@foolip does this make sense to you?

MediaRecorder.html
```diff
@@ -662,6 +662,7 @@
         [Constructor (DOMString type, BlobEventInit eventInitDict)]
         interface BlobEvent : Event {
           readonly attribute Blob data;
```
foolip: Off topic: this could be `[SameObject]`.

MediaRecorder.html
```diff
@@ -712,12 +713,19 @@
...
+          <dd>The encoded blob. Its
+          <a href="https://www.w3.org/TR/FileAPI/#dfn-type">
+          type</a> attribute correlates to the selected
+          <a href="#dom-mediarecorder-mimeType">mimeType</a>.</dd>
+
+          <dt><dfn><code>timecode</code></dfn> of type <span class "idlAttrType">
+          <a href="https://www.w3.org/TR/hr-time-2/#dom-domhighrestimestamp">
+          DOMHighResTimeStamp</a></span>, readonly</dt>
+          <dd><code>Timecode</code> associated with the Blob data, relative to
+          the first produced <code>blob</code>.</dd>
```
foolip: Can the first timecode be anything, or must it be zero? I'm thinking about what tests could be written based on this prose alone.

foolip:
Seems sensible, but another prospective implementer should also review.

MediaRecorder.html
yell0wd0g: In the spirit of RTP, I would just leave it unspecified, the only requirement being that successive `timecode`s increase ( not necessarily monotonically), but I can be convinced otherwise.

yell0wd0g:
@martinthomson, @mreavy could either of you take a look at this PR and/or forward to interested peeps? Thx.

MediaRecorder.html

yell0wd0g:
@jesup, @jan-ivar, WDYT about this proposal?

MediaRecorder.html
```diff
@@ -661,7 +661,8 @@
       <pre class="idl">
         [Constructor (DOMString type, BlobEventInit eventInitDict)]
         interface BlobEvent : Event {
-          readonly attribute Blob data;
+          [SameObject] readonly attribute Blob data;
+          [SameObject] readonly attribute DOMHighResTimeStamp timecode;
```
jan-ivar: Can't put [SameObject] on a double AFAIK.

jan-ivar: Also, should probably add it to BlobEventInit as well, to keep up with the copy-construction pattern?

MediaRecorder.html
```diff
@@ -712,12 +713,19 @@
...
-          href="https://www.w3.org/TR/FileAPI/#blob"><code>Blob</code></a>
-          object whose type attribute indicates the encoding of the blob
-          data.</dd>
+          <dt><dfn><code>data</code></dfn> of type <span class="idlAttrType">
+          <a href="https://www.w3.org/TR/FileAPI/#blob">Blob</a></span>,
+          readonly</dt>
+          <dd>The encoded blob. Its
+          <a href="https://www.w3.org/TR/FileAPI/#dfn-type">
+          type</a> attribute correlates to the selected
+          <a href="#dom-mediarecorder-mimeType">mimeType</a>.</dd>
```
jan-ivar: This link is not working for me.

MediaRecorder.html
jan-ivar: The < code >  formatting in this sentence seems inconsistent with the rest of the doc (elsewhere this spec says "Blob"). Also, timecode being an uncommon word, a plain-text rewording to avoid a self-referential definition here seems preferable, or link to some definition if there is one.

jan-ivar: Also, while Chrome appears to fire `dataavailable` for every frame, this isn't actually to spec. A blob can contain many frames, even the whole recording. The latter is in fact the default, unless [timeslice](https://w3c.github.io/mediacapture-record/MediaRecorder.html#dom-mediarecorder-start) is specified.

What does timecode mean in that case?

jan-ivar: Could you elaborate on why a `DOMHighResTimeStamp` is needed (not saying it isn't, I'd just like to know what the rationale is).

Its [definition](https://www.w3.org/TR/hr-time-2/#idl-def-domhighrestimestamp) says this value is usually a stamp relative to a well-defined [time origin](https://www.w3.org/TR/hr-time-2/#time-origin) (not true here from what I've read), OR a delta between two `DOMHighResTimeStamp`s. If we're going to use this we should at least point out that it's the latter.

Also, why call it timecode and not timestamp? Is there a good reason or precedent?

MediaRecorder.html
yell0wd0g: Oops, should be all lowercase.

MediaRecorder.html
yell0wd0g: Yeah, `Timecode` should not be in a `<code>` tag. Also, it should be 
<var>blob</var> (==`<var>blob</var>`)
, or perhaps 
<a>Blob</a> <var>blob</var> (== `<a>Blob</a> <var>blob</var>.`)

yell0wd0g: > What does timecode mean in that case?

If the recorded blob contains a lot of stuff that could have independent `timecode`s, I'd say the `BlobEvent` `timecode` should be the earliest one.

> Why `DomHighResTimestamp`? 

The alternative [`DOMTimestamp`](https://www.w3.org/TR/DOM-Level-3-Core/core.html#DOMTimeStamp) is just a number of milliseconds in a `long long` format, whereas [`DomHighResTimestamp`](https://www.w3.org/TR/hr-time-2/#dom-domhighrestimestamp) conveys the meaning of time measured from a given origin (and, is `double`, which is good).

I think we should specify that the `origin` in this case is the first `Blob` produced by MR (but we don't have to specify this value to be zero).

`timecode` comes from the film industry: https://en.wikipedia.org/wiki/Timecode

jan-ivar: Sounds good. We can probably just specify it represents the delta between the earliest frame in this blob, and the earliest frame in the first blob. That way we're compliant with the second definition in [DomHighResTimestamp](https://www.w3.org/TR/hr-time-2/#dom-domhighrestimestamp), and we avoid any reference to *time origin* in the first definition, which seems to have a specific definition incompatible with our use here.

yell0wd0g: Took another stab at writing this, ptal

MediaRecorder.html
```diff
@@ -712,12 +713,26 @@
...
+          <a href="https://www.w3.org/TR/hr-time-2/#dom-domhighrestimestamp">
+          DOMHighResTimeStamp</a></span>, readonly</dt>
+          <dd>The difference between the current <var>data</var>'s timestamp and
+          the <var>timecode</var> of the first produced <a>BlobEvent</a>.
+          If <var>data</var> is composed of several chunks that could have
+          independent <var>timecode</var>s (e.g. several encoded video frames),
+          then <var>timecode</var> MUST refer to the earliest individually
+          addressable chunk.
+          Note that the <var>timecode</var> in the first produced
+          <a>BlobEvent</a> does not need to be zero.
```
yell0wd0g, line 723: @jan-ivar , does the current writing lgty? If so, I'll merge.

MediaRecorder.html
```diff
@@ -724,14 +724,14 @@
           <dt><dfn><code>timecode</code></dfn> of type <span class "idlAttrType">
           <a href="https://www.w3.org/TR/hr-time-2/#dom-domhighrestimestamp">
           DOMHighResTimeStamp</a></span>, readonly</dt>
-          <dd>Timecode associated with <var>data</var>.
-          The timestamp MUST be relative to the first produced
-          <a>BlobEvents</a>'s <a>timecode</a> (note that this value does not
-          need to be zero).
+          <dd>The difference between the current <var>data</var>'s timestamp and
+          the <var>timecode</var> of the first produced <a>BlobEvent</a>.
```
jan-ivar: s/timecode/timestamp/ here, otherwise the definition seems recursive and wrong (isn't first timecode ~0)?

I think the terminology here has to be that chunks have timestamps (relative to some common origin of no importance) and that`timecode = timestampB - timestampA`. Do I have the intent right?

MediaRecorder.html
```diff
@@ -724,14 +724,14 @@
...
           If <var>data</var> is composed of several chunks that could have
-          independent timecodes (e.g. several encoded video frames), then
-          <code>timecode</code> MUST refer to the earliest individually
+          independent <var>timecode</var>s (e.g. several encoded video frames),
+          then <var>timecode</var> MUST refer to the earliest individually
```
jan-ivar: Same here. Would it be easier to fold this into the first sentence?

"The difference between the timestamp of the first chunk in *data* and the timestamp of the first chunk in the first BlobEvent produced by this recorder." ?

MediaRecorder.html
```diff
@@ -724,14 +724,14 @@
...
           addressable chunk.
+          Note that the <var>timecode</var> in the first produced
+          <a>BlobEvent</a> does not need to be zero.
```
jan-ivar, line 723: Doesn't it have to be from this definition? Maybe I've misunderstood something.

yell0wd0g: The first timecode is taken as a reference, later ones refer to it. It could be zero, or it could be a OS timestamp, or some random thing like RTP does.

MediaRecorder.html
yell0wd0g: Yes, that's my understanding too.

Will do s/timecode/timestamp/

MediaRecorder.html
yell0wd0g: Done rewritting.

yell0wd0g:
@jan-ivar  ptal

yell0wd0g:
ping @jan-ivar , does it read good to you now?
(I push-forced the changes due to some mysterious rebase error, contents are straightforward to review anyway).

jan-ivar (approved):
lgtm

Issue #99, opened 2016-12-07
ignoreMutedMedia on a combination of muted and unmuted tracks?


alvestrand:
If a stream contains two audio tracks, the "ignoreMutedMedia" flag is set, and one (and only one) track is muted, should recording of the other one continue or not?
At the moment, the text under the "start" method is unclear about this.

yell0wd0g:
Will clarify.  Note that Gecko doesn't have this flag (@jesup, @jan-ivar): https://github.com/mozilla/gecko-dev/blob/86897859913403b68829dbf9a154f5a87c4b0638/dom/webidl/MediaRecorder.webidl

jan-ivar:
Also, what's the behavior if I do this?
```js
let recorder = new MediaRecorder(stream);
setInterval(() => recorder.ignoreMutedMedia = !recorder.ignoreMutedMedia, 2000);
setInterval(() => stream.getTracks().forEach(t => t.enabled = !t.enabled), 3000);
```

yell0wd0g:
The first repeating task will flip the value of `ignoreMutedMedia` from true to false. According to the current reading of the spec, MR will either ignore a muted track (`ignoreMutedMedia` == true) or insert blanks in the recording at the speed of the source (`ignoreMutedMedia` == false).

But, IIRC: `MediaStreamTrack.muted` is a read-only attribute that doesn't relate to `enabled` (*), so the second line doesn't affect the `muted` value, and so renders the behaviour of the lines @jan-ivar posted totally dependent on the behaviour of `stream` :smile:


(*) In [Chromium](https://cs.chromium.org/chromium/src/content/renderer/media/video_track_adapter.cc?sq=package:chromium&dr=CSs&rcl=1481717748&l=450), video `muted` state is updated by monitoring the amount of frames passing versus the frame rate.

https://w3c.github.io/mediacapture-record/MediaRecorder.html#dom-mediarecorder-ignoremutedmedia

jan-ivar:
(Note I forgot to add `rec.start();` to my example).

Hmm, was it really the intent of this spec to respect only the `muted` state and not the `enabled` state? Both do the same thing (inversely) - causing black/silence to be output - and differ only in that the former is out of control of the application. Seems like a mistake to me. What was the intent?

And my question is the same for `muted` state: What's the behavior if recording has started and a track vacillates between muted and unmuted say every 2 seconds? Does it get recorded or not?

[...]

PR #104, opened 2016-12-14
Rewritten |ignoreMutedMedia| parts (#99)

yell0wd0g:


yell0wd0g:
@alvestrand  PTAL

MediaRecorder.html
```diff
@@ -326,11 +329,11 @@
...
-            <p> If any Track within the MediaStream is muted at any time (i.e.,
-            if its <code>muted</code> flag is set to <code>true</code>), the UA
-            MUST insert black frames or silence until the Track is unmuted, or
-            record nothing, depending on the value of
-            <a>ignoreMutedMedia</a>.</p>
+            <p> The UA MUST insert black frames or silence, or record nothing,
+            depending on the value of <a>ignoreMutedMedia</a>, for as long as
+            a given <var>track</var> within the MediaStream  is
+            <a href="https://www.w3.org/TR/mediacapture-streams/#widl-MediaStreamTrack-muted">
+            muted</a>. This behaviour only applies to the muted tracks.</p>
```
alvestrand, line 337: This still does not address the question of what happens if you do

getUserMedia(audio: true, video:true).then(stream => stream.getVideoTracks()[0].muted = true)

and record the result.

Suggested rephrase:

If all tracks are muted, and the "ignoreMutedMedia" flag is set to "true", record nothing (stop adding data to the recording).
Otherwise, the UA MUST insert black frames or silence until the track is unmuted.

yell0wd0g: `muted` is a read-only attribute, so the code would not work.  But more importantly, I think `ignoreMutedMedia` works on a per-track basis, IOW, if MR is recording, say, 3 tracks, and one of them becomes `muted`, then a) the other two keep recording unperturbed and b) what MR records for the `muted` track depends on `ignoreMutedMedia`.

alvestrand: In a multitrack recording, is it possible to record nothing on one of the tracks?
I can't figure out what it means in terms of timestamps; if track A is recorded and track B is muted for 10 seconds with ignoreMutedMedia=true, will they then be 10 seconds out of sync?

WRT setting it, I was thinking of "enabled=false", which is JS-settable. Is "enabled" interacting with recording at all, or does it just cause blackness?

yell0wd0g: >  is it possible to record nothing on one of the tracks?
 
Yes, it'd be possible. Every recorded video frame has a timestamp associated, if there is a jump in any of the sequences of timestamps, any player would just keep the last drawn canvas, unless, that is, `ignoreMutedMedia` inserts black video frames. 

In the hypothetical case you mention, assuming A and B are video tracks for simplicity, after the `muted` hiatus in track B, they'll still be in sync, because the UA will still produce correct timestamps in the container. If you were playing both in parallel, then A will play all the way with no hiccups, whereas B will either be frozen for 10 seconds, or turn black for ten seconds, depending on `ignoreMutedMedia`.

I'm not sure about the extent of `enabled` versus `muted`. In Chrome, `muted` is a reflection of no video passing by the track during a certain period of time; this being affected by `enabled` or not, depends on where the video flow is interrupted (before or after the Track). Maybe @guidou knows more?

jan-ivar: @alvestrand I also reacted to the reference being to `muted` only and not `enabled` as well. Seems like a mistake to me.

yell0wd0g:
If #105 goes ahead removing `ignoreMutedMedia`, I'll nuke this PR.

PR #105, opened 2016-12-17
Remove |ignoreMutedMedia| and references to it. (#99)

yell0wd0g:


MediaRecorder.html
```diff
@@ -326,12 +313,6 @@
             playable, but the combination of all the Blobs from a completed
             recording MUST be playable.</p>
 
-            <p> If any Track within the MediaStream is muted at any time (i.e.,
```
alvestrand, line 329: I think you should keep the "record black and silence" text, just to be explicit.

alvestrand (approved):
Good resolution. Suggest leaving in "record black and silence" for clarity.

jan-ivar (approved):
Lgtm, but I agree with @alvestrand. Also I think we meant `muted` or disabled  (`!enabled`) here FWIW (unsure how to write that).

MediaRecorder.html
yell0wd0g: Rewritten

yell0wd0g:
I restated the previously removed paragraph saying that only black frames and/or silence will be recorded, since that's what a Track should produce when `muted` or `disabled` (https://www.w3.org/TR/mediacapture-streams/#track-enabled) : 
> the consumer gets zero-information-content, which means silence for audio and black frames for video.

Although I believe that's not what Chromium implementation does: instead of black frames or silence, data sinks will get nothing at all if the track is muted or disabled.  But this pertains to implementation bugs.

Issue #53, opened 2016-04-13
Error member in an Event for runtime-errors is a bad idea


jan-ivar:
...because the `Error`'s file, line number and stack will always be empty.

Not only is this wasteful, it fools people into thinking it _might_ have useful information that they better log or forward, taking the place of more useful information.

Here's an [example](https://jsfiddle.net/tmwLxjLy/) (yes I like promises, and yes the example is mediaRecorder,  but it has the same event):

``` js
var rec = new MediaRecorder(stream);
rec.start();
var stopped = new Promise((resolve, reject) => {
  rec.onstop = resolve;
  rec.onerror = event => reject(event.error));
});
```

`stopped` resolves when the recording ends, or rejects at the first runtime error, whichever happens first.

But when a `SecurityError` error later shows up in my console, there's no file or line number. In a larger program, this would suck.

If we replace `.error` with `.name` and `.message` then I know there's no deeper information to forward:

```
rec.onerror = event => reject(new DOMException(event.message, event.name));
```

And now the `SecurityError` points to this call site, which tells me it's a runtime error in `rec`.

I think this is a general problem with all runtime-error events.

jan-ivar:
This issue is a clone of https://github.com/w3c/mediacapture-main/issues/337#issuecomment-209468754 (closed because mediacapture didn't need the error event in the end, but still relevant here).

yell0wd0g:
Merging #31 into this issue.  

The Spec currently defines an [`ErrorEvent`](https://w3c.github.io/mediacapture-record/MediaRecorder.html#errorevent) that just acts as a holder for a `DOMException`.

Gecko defines a RecordErrorEvent [1] that is used in MediaRecorder.cpp [2] by filling its name in NotifyError() to e.g. SecurityError, OutOfMemoryError or otherwise GenericError.  Chromium has no specific error.

Coalescing other conversations:
- using [`ErrorEvent`](https://html.spec.whatwg.org/multipage/webappapis.html#the-errorevent-interface) is not a good idea here because of the `filename`, `lineno` and `colno` that hints as to it being an error produced in a file;
- we still would  like to produce an Error Event interface with a `message` field;
- I personally think it'd be an overkill to define a sort-of `MediaRecorderErrorEvent` just to add a `DOMString` message field.

@jan-ivar, @foolip, what would you suggest we do?


[1] https://github.com/mozilla/gecko-dev/blob/master/dom/webidl/RecordErrorEvent.webidl
[...]

PR #111, opened 2016-12-22
`MediaRecorderErrorEvent`

yell0wd0g:
First go at cleaning up `ErrorEvent`: 
- renamed it to `MediaRecorderErrorEvent` and also s/`ErrorEventInit`/`MediaRecorderErrorEventInit`/
- the Init dict holds now a DOMException, so its use in the ctor is just as a copy-constructor.

### DIFF
--- a/MediaRecorder.html
+++ b/MediaRecorder.html
@@ -779,21 +779,27 @@
       <p>The <a>MediaRecorderErrorEvent</a> interface is defined for cases when
       an event is raised that could have been caused by an error.</p>
       <p><dfn data-dfn-type="dfn" data-lt="firing an error event">To fire an
-      error event</dfn> named <var>e</var> with a <code>DOMException</code> named <var>error</var> means
-      that an event with the name <var>e</var>, which does not bubble (except
-      where otherwise stated) and is not cancelable (except where otherwise
-      stated), and which uses the <code><a>MediaRecorderErrorEvent</a></code>
-      interface with the <code><a data-link-for= "MediaRecorderErrorEvent">error
-      </a></code> attribute set to <var>error</var>, must be created and
+      error event</dfn> named <var>e</var> with a <code>DOMException</code>
+      named <var>error</var> means that an event with the name <var>e</var>,
+      which does not bubble (except where otherwise stated) and is not
+      cancelable (except where otherwise stated), and which uses the
+      <code><a class="idlType">MediaRecorderErrorEvent</a></code> interface with
+      the <code><a data-link-for= "MediaRecorderErrorEvent">error</a></code>
+      attribute set to <var>error</var>, must be created and
       <a href="https://dom.spec.whatwg.org/#concept-event-dispatch">dispatched
       </a> at the given target. If no <code>DOMException</code> object is
       specified, the <code><a data-link-for="MediaRecorderErrorEvent">error
       </a></code> attribute defaults to null.</p>
+
       <div>
       <pre class="idl">
-        [Exposed=Window, Constructor (DOMString name, optional DOMString message)]
+        dictionary MediaRecorderErrorEventInit : EventInit {
+          required DOMException error;
+        };
+
+        [Exposed=Window, Constructor (DOMString type, MediaRecorderErrorEventInit eventInitDict)]
         interface MediaRecorderErrorEvent : Event {
-          [SameObject] readonly attribute DOMException error;
+          [SameObject] readonly attribute DOMException? error;
         };
       </pre>
       <section>
@@ -813,27 +819,23 @@
               <th>Description</th>
             </tr>
             <tr>
-              <td class="prmName">name</td>
+              <td class="prmName">type</td>
               <td class="prmType"><code>DOMString</code></td>
               <td class="prmNullFalse"><span role="img" aria-label="False">&#10008;</span></td>
               <td class="prmOptFalse"><span role="img" aria-label="False">&#10008;</span></td>
               <td class="prmDesc">
               The
               <a href="https://heycam.github.io/webidl/#idl-DOMException-error-names">
-              error name</a> to be used to construct
+              error name</a> to be used in constructing
               <a data-link-for="MediaRecorderErrorEvent">error</a>.
               </td>
             </tr>
             <tr>
-              <td class="prmName">message</td>
-              <td class="prmType"><code>DOMString</code></td>
+              <td class="prmName">eventInitDict</td>
+              <td class="prmType"><a class="idlType"><code>MediaRecorderErrorEventInit</code></a></td>
               <td class="prmNullFalse"><span role="img" aria-label="False">&#10008;</span></td>
               <td class="prmOptTrue"><span role="img" aria-label="False">&#10004;</span></td>
-              <td class="prmDesc">An explanatory message about the error
-              circumstances. This argument, if present, will be used to
-              initialize <code><a data-link-for="MediaRecorderErrorEvent">
-              error</a></code>'s <a href="https://www.w3.org/TR/WebIDL-1/#es-DOMException-constructor-object">message</a>.
-              </td>
+              <td class="prmDesc"></td>
             </tr>
           </tbody>
           </table>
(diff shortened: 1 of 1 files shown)


yell0wd0g:
@jan-ivar, I'm not sure if this is closer to what you wanted, or farther away in a diametrically opposed fashion (because I moved away from `ErrorEvent`, so there won't be any info on `line` etc).  In the interest of review speed, and if we are in the second case, can I ask you to provide concrete wordings of what you'd like to see written in this Section?

MediaRecorder.html
```diff
@@ -775,97 +775,101 @@
...
-      data-lt="firing an error event">To fire an error event</dfn> named
-      <var>e</var> with a <code>DOMException</code> named <var>error</var> means
-      that an event with the name <var>e</var>, which does not bubble (except
-      where otherwise stated) and is not cancelable (except where otherwise
-      stated), and which uses the <code><a>ErrorEvent</a></code> interface with
-      the <code><a data-link-for= "ErrorEvent">error</a></code> attribute set to
-      <var>error</var>, must be created and
+      <h3>MediaRecorderErrorEvent</h3>
+      <p>The <a>MediaRecorderErrorEvent</a> interface is defined for cases when
+      an event is raised that could have been caused by an error.</p>
```
jan-ivar: We can say "was caused by" now.

MediaRecorder.html
```diff
@@ -775,97 +775,101 @@
...
-          "attributes">
-            <dt><dfn><code>error</code></dfn> of type <span class=
-            "idlAttrType"><a>Error</a></span>, readonly , nullable</dt>
-            <dd>If the event was raised because of an error, this attribute MAY
-            be set to that error object.</dd>
-          </dl>
-        </section>
-      </div>
+      specified, the <code><a data-link-for="MediaRecorderErrorEvent">error
+      </a></code> attribute defaults to null.</p>
```
jan-ivar: error is no longer optional so this is not possible.

MediaRecorder.html
```diff
@@ -775,97 +775,101 @@
...
-        </section>
-      </div>
+      <pre class="idl">
+        dictionary MediaRecorderErrorEventInit : EventInit {
+          required DOMException error;
+        };
+
+        [Exposed=Window, Constructor (DOMString type, MediaRecorderErrorEventInit eventInitDict)]
+        interface MediaRecorderErrorEvent : Event {
+          [SameObject] readonly attribute DOMException? error;
```
jan-ivar: *error* can no longer be `null` since it is required in the constructor, so this should not be nullable anymore.

MediaRecorder.html
```diff
@@ -775,97 +775,101 @@
...
+              <a href="https://heycam.github.io/webidl/#idl-DOMException-error-names">
+              error name</a> to be used in constructing
+              <a data-link-for="MediaRecorderErrorEvent">error</a>.
+              </td>
+            </tr>
+            <tr>
+              <td class="prmName">eventInitDict</td>
+              <td class="prmType"><a class="idlType"><code>MediaRecorderErrorEventInit</code></a></td>
+              <td class="prmNullFalse"><span role="img" aria-label="False">&#10008;</span></td>
+              <td class="prmOptTrue"><span role="img" aria-label="False">&#10004;</span></td>
```
jan-ivar: prmOptFalse

MediaRecorder.html
```diff
@@ -775,97 +775,101 @@
...
+        </dl>
+      </section>
+      <section>
+        <h2>Attributes</h2>
+        <dl data-link-for="MediaRecorderErrorEvent"
+            data-dfn-for="MediaRecorderErrorEvent" class="attributes">
+        <dt><dfn><code>error</code></dfn> of type <span class="idlAttrType">
+          <a href="https://heycam.github.io/webidl/#dfn-DOMException">
+          DOMException</a></span>, readonly</dt>
+        <dd>The DOMException error that triggered the event, if any.</dd>
```
jan-ivar: s/, if any//

MediaRecorder.html
```diff
@@ -775,97 +775,101 @@
...
+        </dl>
+      </section>
+      <section>
+        <h2>Dictionary <a class="idlType">MediaRecorderErrorEventInit</a> members</h2>
+        <dl data-link-for="MediaRecorderErrorEventInit"
+            data-dfn-for="MediaRecorderErrorEventInit" class="dictionary-members">
+        <dt><dfn><code>error</code></dfn> of type <span class="idlAttrType">
+          <a href="https://heycam.github.io/webidl/#dfn-DOMException">
+          DOMException</a></span>, readonly</dt>
```
jan-ivar: s/readonly/required/

MediaRecorder.html
```diff
@@ -775,97 +775,101 @@
...
+        <dd>The <a href="https://heycam.github.io/webidl/#dfn-DOMException">
+          DOMException</a> causing the error that triggered the event.
+          An explanatory message about the error circumstances MAY be provided
+          in its <a href="https://www.w3.org/TR/WebIDL-1/#es-DOMException-constructor-object">
```
jan-ivar: This paragraph links to two different definitions of `DOMException`. We should pick one.

jan-ivar (changes requested):
Remove nullable, and misc nits. Otherwise looks good!

jan-ivar:
Fix for https://github.com/w3c/mediacapture-record/issues/53.

MediaRecorder.html
```diff
@@ -775,97 +775,101 @@
...
+      <section>
+        <h2>Dictionary <a class="idlType">MediaRecorderErrorEventInit</a> members</h2>
+        <dl data-link-for="MediaRecorderErrorEventInit"
+            data-dfn-for="MediaRecorderErrorEventInit" class="dictionary-members">
+        <dt><dfn><code>error</code></dfn> of type <span class="idlAttrType">
+          <a href="https://heycam.github.io/webidl/#dfn-DOMException">
+          DOMException</a></span>, readonly</dt>
+        <dd>The <a href="https://heycam.github.io/webidl/#dfn-DOMException">
+          DOMException</a> causing the error that triggered the event.
+          An explanatory message about the error circumstances MAY be provided
```
jan-ivar, line 832: @miguelao @domenic Maybe we could add something here like

>  **Note:** If an implementation places non-standard properties on `DOMException`, exposing for example stack traces or error line numbers, these are encouraged to point to whichever method call most closely identifies the run-time operation that experienced the error. e.g. `start()`.

MediaRecorder.html
```diff
@@ -775,97 +775,101 @@
...
+          required DOMException error;
+        };
+
+        [Exposed=Window, Constructor (DOMString type, MediaRecorderErrorEventInit eventInitDict)]
+        interface MediaRecorderErrorEvent : Event {
+          [SameObject] readonly attribute DOMException? error;
+        };
+      </pre>
+      <section>
+        <h2>Constructors</h2>
```
domenic, line 804: You should not specify a constructor for events. They all use the same constructor: https://dom.spec.whatwg.org/#concept-event-constructor

MediaRecorder.html
domenic: https://heycam.github.io/webidl/#dfn-DOMException please; it has several important fixes

MediaRecorder.html
domenic: SGTM

MediaRecorder.html

MediaRecorder.html

MediaRecorder.html
yell0wd0g: Done removing `?`

MediaRecorder.html

MediaRecorder.html

MediaRecorder.html

MediaRecorder.html
yell0wd0g: Changed to pointing to the constructor in https://heycam.github.io/webidl/#es-DOMException-constructor-object (ISO the one in w3.org).

MediaRecorder.html
yell0wd0g: Done adding the proposed note.

MediaRecorder.html
yell0wd0g: Removed the table and added the link you mention, but left the constructor in the idl.  Or should I remove all references to the ctor?

The two examples I found: https://www.w3.org/TR/mediacapture-streams/#properties-of-overconstrainederror-instances and https://w3c.github.io/clipboard-apis/#clipboard-event-interfaces both explicit the constructor in the idl.

jan-ivar (approved):
Lgtm.

PR #118, opened 2017-02-14
Make Error Event go after ondataavailable.

### LINKED ISSUE #52: Ensure that the error event is the last event to be fired
http://w3c.github.io/mediacapture-record/MediaRecorder.html#general-principles

> If recording has been started and not yet stopped when the error occurs, then after raising the error, the UA must fire a `dataavailable event`, containing any data that it has gathered, and then a stop event.

This seems odd, it will mean that sometimes there will be data after the error event and sometimes there will not. How about instead requiring that if there is any data, that the `dataavailable` event be fired before the `error` event? Assuming that both the data and the error originates from the same bits of code then that ought to also match the actual order things happen in.

yell0wd0g:
Upon an error condition, first fire an Event called `ondataavailable`, then an Event called `stop` and finally fire an error event.

Addresses #52 .

Chrome update in progress in https://crrev.com/2697703003

***
<a href="https://pr-preview.s3.amazonaws.com/yellowdoge/mediacapture-record/pull/118.html" title="Last updated on Feb 15, 2021, 7:32 AM UTC (96db655)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/mediacapture-record/118/faf208b...yellowdoge:96db655.html" title="Last updated on Feb 15, 2021, 7:32 AM UTC (96db655)">Diff</a>

### DIFF
--- a/MediaRecorder.bs
+++ b/MediaRecorder.bs
@@ -286,7 +286,7 @@ interface MediaRecorder : EventTarget {
         <td class="prmNullFalse"><span role="img" aria-label="False">&#10008;</span></td>
         <td class="prmOptTrue"><span role="img" aria-label="True">&#10004;</span></td>
         <td class="prmDesc">The minimum number of milliseconds of data
-        to return in a single Blob.</td>
+        to return in a single {{Blob}}.</td>
       </tr>
     </tbody>
   </table>
@@ -302,7 +302,7 @@ interface MediaRecorder : EventTarget {
     task, using the DOM manipulation task source, that runs the following steps:
     <ol>
       <li>Set {{state}} to {{inactive}} and stop gathering data.</li>
-      <li>Let <var>blob</var> be the Blob of collected data so far and
+      <li>Let <var>blob</var> be the {{Blob}} of collected data so far and
       let <var>target</var> be the MediaRecorder context object, then
       <a href="#to-fire-a-blob-event">fire a blob event</a> named
       <a>dataavailable</a> at <var>target</var> with <var>blob</var>.</li>
@@ -369,7 +369,7 @@ interface MediaRecorder : EventTarget {
       (Note that <var>blob</var> will be empty if no data has been gathered
       yet.)</li>
 
-      <li>Create a new Blob and gather subsequent data into it.</li>
+      <li>Create a new {{Blob}} and gather subsequent data into it.</li>
     </ol>
     </li>
     <li>return <code>undefined</code>.</li>
@@ -570,31 +570,28 @@ dictionary BlobEventInit {
 
 ## General principles ## {#error-handling-principles}
 
-If at any moment the UA finds an error condition, the UA MUST run the following
-steps:
-  <ol>
-    <li>If {{state}} is {{inactive}} throw an {{InvalidStateError}}
-    {{DOMException}} and terminate these steps. Otherwise the UA MUST queue a
-    task, using the DOM manipulation task source, that runs the following steps:
-    <ol>
-      <li>Set {{state}} to {{inactive}} and stop gathering data.</li>
-      <li>Let <var>blob</var> be the {{Blob}} of collected data so far and
-      let <var>target</var> be the MediaRecorder context object</li>
+If at any moment the UA finds an error condition, the UA MUST queue a task,
+using the DOM manipulation task source, that runs the following steps:
+<ol>
+  <li>Set {{state}} to {{inactive}} and stop gathering data.</li>
 
-      <li><a href="#to-fire-a-blob-event">Fire a blob event</a> named
-        <a>dataavailable</a> at <var>target</var> with <var>blob</var>.</li>
-      <li><a>Fire an event</a> named <a>stop</a> at <var>target</var></li>
-      <li><a>Fire an error event</a> named {{MediaRecorderErrorEvent}} at
-        <var>target</var>.</li>
-    </ol>
-  </li>
+  <li>Let <var>blob</var> be the {{Blob}} of collected data so far and
+  let <var>target</var> be the MediaRecorder context object</li>
+
+  <li><a href="#to-fire-a-blob-event">Fire a blob event</a> named
+  <a>dataavailable</a> at <var>target</var> with <var>blob</var>.</li>
+
+  <li><a>Fire an event</a> named <a>stop</a> at <var>target</var></li>
+
+  <li><a>Fire an error event</a> named <var>event</var> at
+  <var>target</var>.</li>
 </ol>
 
 <div class="note">
-  The UA may set platform-specific limits, such as those for the minimum and
-  maximum {{Blob}} size that it will support, or the number of
-  {{MediaStreamTrack}}s it will record at once. It will signal a fatal error if
-  these limits are exceeded.
+  The UA MAY set platform-specific limits, e.g. for the minimum and maximum
+  {{Blob}} size that it will support, or the number of {{MediaStreamTrack}}s it
+  will record at once. It will signal a fatal error if these limits are
+  exceeded.
 </div>
 
 ## MediaRecorderErrorEvent ## {#errorevent-section}
--- a/MediaRecorder.bs
+++ b/MediaRecorder.bs
@@ -569,20 +569,33 @@ dictionary BlobEventInit {
 # Error handling # {#error-handling}
 
 ## General principles ## {#error-handling-principles}
-<em>This section is non-normative.</em>
 
-The UA will throw a {{DOMException}} when the error can be detected at the time
-that the call is made. In all other cases the UA will <a>fire an event</a> named
-{{MediaRecorderErrorEvent}}.  If recording has been started and not yet stopped
-when the error occurs, let <var>blob</var> be the {{Blob}} of collected data so
-far; before raising the error, the UA will <a href="#to-fire-a-blob-event">fire a
-dataavailable event</a> with <var>blob</var>; then the UA will
-<a>fire an event</a> named <code>stop</code>; finally the UA will
-<a>fire an error event</a>.
-The UA may set platform-specific limits, such as those for the minimum and
-maximum {{Blob}} size that it will support, or the number of
-{{MediaStreamTrack}}s it will record at once.
-It will signal a fatal error if these limits are exceeded.
+If at any moment the UA finds an error condition, the UA MUST run the following
+steps:
+  <ol>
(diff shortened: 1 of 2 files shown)


yell0wd0g:
@foolip tiny change to write what we agreed upon in the bug, ptal.

MediaRecorder.bs
```diff
@@ -575,9 +575,10 @@ The UA will throw a {{DOMException}} when the error can be detected at the time
 that the call is made. In all other cases the UA will <a>fire an event</a> named
 {{MediaRecorderErrorEvent}}.  If recording has been started and not yet stopped
 when the error occurs, let <var>blob</var> be the {{Blob}} of collected data so
-far; after raising the error, the UA will <a href="#to-fire-a-blob-event">fire a
-dataavailable event</a> with <var>blob</var>; immediately after the UA will then
-<a>fire an event</a> named <code>stop</code>.
+far; before raising the error, the UA will <a href="#to-fire-a-blob-event">fire a
```
foolip: The use of "will" here makes this sound not normative but rather descriptive. If this is the only place that ensures the order of things, can it say "must" more?

foolip: Also, in normative requirements, I would expect to see something like "queue a task to..." or clearly requiring that something happens synchronously. I'm not sure if the difference is observable, but basically I would expect the words here to map 1:1 to what an implementation actually does.

yell0wd0g: (After some offline chatting, we've decided that is best to beef up this subsection and make it normative with the appropriate writing, otherwise is quite misleading. New PS coming imminently.)

yell0wd0g:
Ready for review @foolip

MediaRecorder.bs
```diff
@@ -569,19 +569,33 @@ dictionary BlobEventInit {
...
-<a>fire an event</a> named <code>stop</code>.
-The UA may set platform-specific limits, such as those for the minimum and
-maximum {{Blob}} size that it will support, or the number of
-{{MediaStreamTrack}}s it will record at once.
-It will signal a fatal error if these limits are exceeded.
+
+If at any moment the UA finds an error condition, the UA MUST run the following
+steps:
+  <ol>
+    <li>If {{state}} is {{inactive}} throw an {{InvalidStateError}}
```
foolip: The error probably happens in the background, so there isn't anywhere to throw the exception. Can there even be an error when inactive? Is it because state becomes inactive before it's *really* inactive?

MediaRecorder.bs
```diff
@@ -569,19 +569,33 @@ dictionary BlobEventInit {
...
+    task, using the DOM manipulation task source, that runs the following steps:
+    <ol>
+      <li>Set {{state}} to {{inactive}} and stop gathering data.</li>
+      <li>Let <var>blob</var> be the {{Blob}} of collected data so far and
+      let <var>target</var> be the MediaRecorder context object</li>
+
+      <li><a href="#to-fire-a-blob-event">Fire a blob event</a> named
+        <a>dataavailable</a> at <var>target</var> with <var>blob</var>.</li>
+      <li><a>Fire an event</a> named <a>stop</a> at <var>target</var></li>
+      <li><a>Fire an error event</a> named {{MediaRecorderErrorEvent}} at
```
foolip: MediaRecorderErrorEvent is the name of the interface, I guess the event type is just "error"?

foolip (approved):
LGTM % nits, if some other implementers takes a look that'd be good.

MediaRecorder.bs
yell0wd0g: Yeah, changed.

MediaRecorder.bs
yell0wd0g: This was some legacy from before where the general principles of error handling stated that we should try to  throw synchronously a `DOMException` if possible at all before having to fire the error event.  The first part is kind-of obvious and is anyway stated individually in each of the commands (`start()`, `stop()` etc), so I'll remove it and go for the meat of the section.

yell0wd0g:
@foolip, thanks, done adding your changes.

@jan-ivar can you take a look plz?

MediaRecorder.bs
```diff
@@ -569,19 +569,30 @@ dictionary BlobEventInit {
...
+
+  <li>Let <var>blob</var> be the {{Blob}} of collected data so far and
+  let <var>target</var> be the MediaRecorder context object</li>
+
+  <li><a href="#to-fire-a-blob-event">Fire a blob event</a> named
+  <a>dataavailable</a> at <var>target</var> with <var>blob</var>.</li>
+
+  <li><a>Fire an event</a> named <a>stop</a> at <var>target</var></li>
+
+  <li><a>Fire an error event</a> named <var>event</var> at
```
foolip: Shouldn't it be "error"? Apart from the name, the *italics* would also cause me to go looking for a variable named *event* which has a string value.

yell0wd0g: Ooops, my bad, I mean `named {{UnknownError}}`, next PS coming.

MediaRecorder.bs
```diff
@@ -569,19 +569,30 @@ dictionary BlobEventInit {
...
+
+  <li>Let <var>blob</var> be the {{Blob}} of collected data so far and
+  let <var>target</var> be the MediaRecorder context object</li>
+
+  <li><a href="#to-fire-a-blob-event">Fire a blob event</a> named
+  <a>dataavailable</a> at <var>target</var> with <var>blob</var>.</li>
+
+  <li><a>Fire an event</a> named <a>stop</a> at <var>target</var></li>
+
+  <li><a>Fire an error event</a> named {{UnknownError}} at <var>target</var>.
```
foolip: OK, now I see why this is confusing. https://w3c.github.io/mediacapture-record/MediaRecorder.html#errorevent-section is an event interface where the member is a DOMException.

The event type (the string that is the first argument to `addEventListener`) should be "error", so this should say 'Fire an error event named "error" with an UnknownError at target'.

I see that the two existing uses of "fire an error event" also has a slightly wrong incantation. It would be possible to just bake the type ("error") into the definition of "fire an error event", but that's an editorial decision.

yell0wd0g: Ok, I'll write something in all three cases like:
`<a>Fire an error event</a> named <a>error</a> with an {{UnknownError}} at    <var>target</var>`

(where `<a>error</a>` links to the `error` entry in the Event Summary section)

That should do the trick IIUC :-)

foolip (approved):

yell0wd0g:
@jan-ivar this should be an easy review, PTAL

yell0wd0g:
@jan-ivar ping

MediaRecorder.bs
```diff
@@ -569,19 +569,30 @@ dictionary BlobEventInit {
...
-far; after raising the error, the UA will <a href="#to-fire-a-blob-event">fire a
-dataavailable event</a> with <var>blob</var>; immediately after the UA will then
-<a>fire an event</a> named <code>stop</code>.
-The UA may set platform-specific limits, such as those for the minimum and
-maximum {{Blob}} size that it will support, or the number of
-{{MediaStreamTrack}}s it will record at once.
-It will signal a fatal error if these limits are exceeded.
+
+If at any moment the UA finds an error condition, the UA MUST queue a task,
+using the DOM manipulation task source, that runs the following steps:
```
jan-ivar, line 574: I worry this is a bit broad.

But more importantly, it looks like we already have normative steps for this as mentioned above, so with this PR there would be two competing normative steps for this condition, which I think is confusing (do we fire twice?) - I'd vote we fix the existing place instead and leave *General Principles* non-normative or remove it, it seems a bit redundant.

MediaRecorder.bs
```diff
@@ -222,8 +222,8 @@ interface MediaRecorder : EventTarget {
       following steps:
       <ol>
         <li>Set {{state}} to {{inactive}}.</li>
-        <li><a>Fire an error event</a> named {{UnknownError}} at
-        <var>target</var>.</li>
+        <li><a>Fire an error event</a> named <a>error</a> with an
+        {{UnknownError}} at <var>target</var>.</li>
```
jan-ivar, line 226: I think we should fix the event order here instead.

It looks like these are the normative steps we already have for recording errors, here in step 5.4 under the [start](https://w3c.github.io/mediacapture-record/MediaRecorder.html#dom-mediarecorder-start) method: *"If the UA at any point is unable to continue gathering data for reasons other than isolation properties or stream track set, it MUST stop gathering data, and queue a task, using the DOM manipulation task source, that runs the following steps"*.

MediaRecorder.bs
```diff
@@ -569,19 +569,30 @@ dictionary BlobEventInit {
...
+
+  <li>Let <var>blob</var> be the {{Blob}} of collected data so far and
+  let <var>target</var> be the MediaRecorder context object</li>
+
+  <li><a href="#to-fire-a-blob-event">Fire a blob event</a> named
+  <a>dataavailable</a> at <var>target</var> with <var>blob</var>.</li>
+
+  <li><a>Fire an event</a> named <a>stop</a> at <var>target</var></li>
+
+  <li><a>Fire an error event</a> named <a>error</a> with an {{UnknownError}} at
```
jan-ivar, line 586: Question: Do we really want to fire `stop` before `error`? It would make it harder to detect errors IMHO. Most code I've seen seem to treat reaching `stop` without `error` as success, and this would break that assumption.

How about this order: `dataavailable`, `error`, `stop`?

jan-ivar (changes requested):
So sorry for missing this request for so long! Hopefully it is not too late to provide feedback—thanks @pehrsons for bringing to my attention.

I agree `dataavailable` should fire before `error`, unsure about `stop`. Also, we seem to have normative steps already that need to be fixed.

alvestrand:
@yellowdoge is this documenting the way Chrome works now after https://crbug.com/693179 was closed (2017)?

jan-ivar:
@Pehrsons do you know if Firefox works the same as Chrome here? Any objection to merging this?

Pehrsons:
Firefox fires events [in this order](https://searchfox.org/mozilla-central/rev/b579290e6b830d1b3f0a941203b0c0e1e56c42a3/dom/media/MediaRecorder.cpp#959,964,979,989):
- "start"
- "dataavailable" unless there is a security error coming
- "error"
- "stop"

My only concern with swapping "error" and "stop" is that it could seem unclear which session the "error" comes from since there potentially could be another "start" in flight at this point.

Issue #133, opened 2017-08-31
Should start() fire the start event before data is received by the recorder?


SingingTree:
The current wording of `start()` includes the following:

> Once media becomes available from one or more of the stream's tracks, start gathering the data into a Blob blob and queue a task, using the DOM manipulation task source, to fire an event named start at target. 

This suggests that the start event should not be fired until data is received. However, current implementations appear to differ in how they behave. When recording from an empty stream Chromium looks to immediately fire the start event, while Firefox will not. [Fiddle of this example](https://jsfiddle.net/SingingTree/arh7t487/).

Is there a preferred approach? Is there a function to waiting until data has arrived before firing start?

SingingTree:
https://github.com/w3c/mediacapture-record/pull/135 being merged answers this: start is fired immediately.

PR #135, opened 2017-09-06
Fire the start event immediately upon start() call

SingingTree:
Related to https://github.com/w3c/mediacapture-record/issues/133. Update spec to state start event should immediately be fired upon a call to start() rather than waiting for some data to be gathered into blob.

ping @jan-ivar

### DIFF
--- a/MediaRecorder.bs
+++ b/MediaRecorder.bs
@@ -197,10 +197,9 @@ interface MediaRecorder : EventTarget {
 
     <li>Set {{state}} to {{recording}}, and run the following steps in parallel:
     <ol>
-      <li>Once media becomes available from one or more of the
-      {{MediaRecorder/stream}}'s tracks, start gathering the data into a
-      {{Blob}} <var>blob</var> and queue a task, using the DOM manipulation task
-      source, to <a>fire an event</a> named <a>start</a> at <var>target</var>.
+      <li>Start gathering the data into a {{Blob}} <var>blob</var> and queue a
+      task, using the DOM manipulation task source, to <a>fire an event</a>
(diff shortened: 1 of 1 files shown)


MediaRecorder.bs
```diff
@@ -257,6 +256,8 @@ interface MediaRecorder : EventTarget {
 
     </ol>
     </li>
+    <li><a>Fire an event</a> named <a>start</a> at <var>target</var>.</li>
```
jan-ivar: So, normally, [in parallel](https://html.spec.whatwg.org/#in-parallel) steps describe steps to be performed in sequence on a background thread. Unfortunately, the approach taken in this method strays from that, in that it describes things that happen later, after the method has returned. That makes it unclear when this event fires.

Short of tearing this structure apart, and moving things to sections outside `start()`, I suggest we instead modify the first step to happen instantly by removing the "Once". Something like:

*"Start gathering data from all of the stream's tracks into a Blob blob, then queue a task, using the DOM manipulation task source, to fire an event named start at target, and then return from this method."*

jan-ivar (changes requested):

SingingTree:
Updated based on feedback from @jan-ivar. Is the language around queuing a task required, or can we fire immediately?

yell0wd0g (approved):
The change LGTM. @jan-ivar WDYT?

jan-ivar (approved):
lgtm.

jan-ivar:
@SingingTree Yes, this is running [in parallel](https://html.spec.whatwg.org/multipage/infrastructure.html#in-parallel), so we must [queue a task](https://html.spec.whatwg.org/multipage/webappapis.html#queue-a-task) to fire the event back on the main thread.

Issue #150, opened 2018-10-08
Not clear if onpause/onresume should not be fired if it is already paused/recording


kbumsik:
Hi, I'm currently fixing Firefox that does not actually fire any onpuase/onresume events.

While fixing code, I noticed that the spec does not make it clear if `pause()`/`resume()` has to simply return `undefined` without firing an event, when its `state` is already `paused`/`recording`? Allowing them fire an event would cause duplicated events.

FYI, Chrome returns undefined at the very beginning when it is already paused/recording state.

alvestrand:
The spec as written clearly says that calling "pause(); pause()" must queue two tasks, both of which fire the "paused" event.

My immediate thought is that this seems silly. Should we change it?
(If we change it, the check for the paused state must be inside the queued task (otherwise race conditions could occur), so we can't let pause() return an error.)

kbumsik:
Allowing only one "paused" event fired makes sense to me too.
> the check for the paused state must be inside the queued task.

I don't know much about implementation details but I'm wondering if it is trivial/common practice? As mentioned in #123, Firefox and Chrome currently change `state` synchronously, [so preventing multiple "paused" event is as easy as adding a single if statement...for now](https://cs.chromium.org/chromium/src/third_party/blink/renderer/modules/mediarecorder/media_recorder.cc?l=278).

Currently Chrome does not fire multiple "paused" event, and Firefox (as my fix to add `onpause`/`onresume` is just accepted by the Firefox devs) fires multiple "paused"/"resume" events. Aside from the state transition being async, if W3C prefers only one "pause" event fired I can submit to patch to Firefox.

Pehrsons:
I agree changing this makes sense. Note that this is affected by #157. CC @jan-ivar

jan-ivar:
This should be quite easy to fix once https://github.com/w3c/mediacapture-record/pull/157 has merged.

jan-ivar:
I'll provide a PR.

jan-ivar:
Actually, I think I need to fix this in https://github.com/w3c/mediacapture-record/pull/157 itself.

yell0wd0g:
Fixed by #157, do we/ should we have a WPT covering this?

jan-ivar:
Thanks @yellowdoge! wpt test is set to arrive from [bug 1514016](https://bugzilla.mozilla.org/show_bug.cgi?id=1514016#c9).

Issue #123, opened 2017-05-12
Should stop(), pause(), resume() change state synchronously?


jan-ivar:
With the exception of [start()](https://w3c.github.io/mediacapture-record/MediaRecorder.html#dom-mediarecorder-start), which sets `state` to `"recording"` synchronously, all the other state-changing methods, stop(), pause(), resume(), set `state` asynchronously, in the queued task that fires the corresponding event.

Do we have a good reason for this? I'm asking because both Chrome and Firefox appear to set `state` synchronously, and fire the event asynchronously. Try it [here](https://jsfiddle.net/jib1/tmufhavj/).

We're [adding `pause` and `resume` events to Firefox](https://bugzil.la/1363915), and aren't sure what to do.

Full disclosure: I fixed the start() method in https://github.com/w3c/mediacapture-record/pull/56. The spec seemed confused about this before then, throwing exceptions in queued tasks.

cc @foolip, @SingingTree

foolip:
When the actual state isn't synchronously available, then I think there are two options for methods and setters that affect state:
* Pretend that the state changes synchronously anyway, and queue a task to fire events. Media elements mostly work like this.
* Instead wait for the real change to complete, and change the state + fire the event at the same time once done. It means one has to deal with different operations racing to complete, or have a queue of operations.

It sounds like there's a mix here, and that it's mostly async. So perhaps best to make everything async. If you ever want these methods to return promises, that'll make more sense too. (When done: change state, resolve promise, fire event.)

yell0wd0g:
So @jan-ivar , @foolip: IIUC the proposal is to make all state transitions asynchronous, which will need modifying the literature of `start()` and adapting the Cr and FF code, correct?

foolip:
I don't have much skin in the game, but yeah, if it seems likely to be web compatible, then going full async seems reasonable.

jan-ivar:
Well, the most web compatible would of course be changing the spec to match Chrome and Firefox. 

Clearly some code would break from this, the question is how much code would break.

For instance, my own [fiddle](http://jsfiddle.net/jib1/tmwLxjLy/) should work, except I apparently do this for debug reasons:
```js
rec.start();
log(rec.state + " for "+ (ms / 1000) +" seconds..."); // recording for 5 seconds
```
which would now say "inactive for 5 seconds" 😜

[...]

PR #157, opened 2018-12-13
stop(), pause() and resume() now set state synchronously, for historical reasons.

jan-ivar:
Fix for https://github.com/w3c/mediacapture-record/issues/123.

### DIFF
--- a/MediaRecorder.bs
+++ b/MediaRecorder.bs
@@ -180,6 +180,10 @@ interface MediaRecorder : EventTarget {
 ## Methods ## {#mediarecorder-methods}
 
 <dl class="domintro">
+  <div class="note">
+    For historical reasons, the following methods alter {{state}} synchronously
+    and fire events asynchronously.
+  </div>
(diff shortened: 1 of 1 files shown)


foolip:
Are there tests in wpt verifying this behavior?

jan-ivar:
@foolip Not yet. We have an action item in [bug 1514016](https://bugzilla.mozilla.org/show_bug.cgi?id=1514016#c9) to provide one.

jan-ivar:
Need to fix https://github.com/w3c/mediacapture-record/issues/150 here at the same time, otherwise `pause()` and `resume()` stop being idempotent.

jan-ivar:
_deleted an incorrect statement_.

yell0wd0g (approved):
I don't see why this shouldn't be merged.

yell0wd0g:
@jan-ivar will let you merge this one yourself seeing that there's another PR on it.

Issue #141, opened 2018-02-16
Add 'privacy and security' section


stefhak:
As described in https://lists.w3.org/Archives/Public/public-media-capture/2018Feb/0005.html this is missing. "MediaStream Recording has no mention of privacy and no privacy or security considerations sections. "

yell0wd0g:
Related to #122

yell0wd0g:
@npdoty  could you follow up and take another look at the current status of the [`Privacy and Security`](https://w3c.github.io/mediacapture-record/#privacy-and-security) section please?

npdoty:
Sure. The section seems especially brief. Have the editors or the group gone through the security questionnaire as described in #122?

* Are there no privacy or security implications beyond use of resources? 
* Any fingerprinting considerations? 
* What scripts have access to the MediaRecorder API? (Parents of an iframe? Child frames? Analytics scripts embedded in the origin?) 
* How are isolation properties of streams handled? 
* Are users provided indication that a stream is being recorded? Could those recordings be communicated to a server or to another origin? How does that compare to a getUserMedia stream where the MediaRecorder API isn't in use?

samuelweiler:
Thanks for tagging this @dontcallmedom.  I see this has been around for a while. Be aware that the current guidelines call for security and privacy sections to be split: https://w3ctag.github.io/security-questionnaire/#considerations

Issue #142, opened 2018-02-16
Fingerprint surface coming from MediaRecorder-able types - needs consideration?


stefhak:
From https://lists.w3.org/Archives/Public/public-media-capture/2018Feb/0005.html

Current specification and implementation also adds an isTypeSupported method which provides drive-by fingerprinting surface, not currently constrained by user permissions or interaction. 
The current draft of our Mitigating Browser Fingerprinting guidance may be helpful: https://w3c.github.io/fingerprinting-guidance/

yell0wd0g:
I don't think [isTypeSupported()](https://w3c.github.io/mediacapture-record/MediaRecorder.html#dom-mediarecorder-istypesupported) increases the fingerprinting: the set of supported profiles is small and has few differences among OSes; it probably never changes among users. E.g. the [implementation status](https://github.com/yellowdoge/mediacapture-record-implementation-status/blob/master/chromium.md) shows which video codecs are implemented in Chrome per platform (there are just a few because Chrome must provide a software fallback in case any hardware-accelerated fail -- this also means that it's not possible for a user to identify those codecs that are hardware accelerated). A malicious site could perhaps infer the OS/platform but that information is already in the User Agent string.

Should this analysis be added to the text?

yell0wd0g:
@npdoty

npdoty:
If the data returned by `isTypeSupported` won't vary between implementations, or there's some assurance that a particular OS/UA implementation won't have variations inside of it, then that would affect the fingerprintability and I do think that's worth documenting in the specification.

The implementation status page you point to specifically notes that the supported values currently differ based on the chipset (ARM or Intel) in Chrome and link to bugs that suggest that the values will differ based on the particular chipset (chipsets that support VP9). Am I reading those tables/bugs correctly? That may be feature detection that's intended for this functionality, but I think we should not claim that there is no variation or information disclosure if that variation is present or planned.

npdoty:
Currently, the first example in the specification demonstrates how to enumerate the list of supported codec/container combinations in the user's browser:
https://w3c.github.io/mediacapture-record/#example1

I don't think you want to recommend doing this kind of enumeration, as it's a particular example of what a site would do in order to fingerprint a user's device rather than detect a capability in order to provide functionality. Some browsers might want to limit the number of successive `isTypeSupported` calls to limit disclosure.

yell0wd0g:
[...]

PR #165, opened 2019-03-08
more "Privacy and Security Considerations" text

yell0wd0g:
(partially) addressing #141 -- arguably a bit delayed, but I didn't know where or how to start, so I read 
https://w3ctag.github.io/security-questionnaire/
https://www.w3.org/TR/fingerprinting-guidance/
and https://www.torproject.org/projects/torbrowser/design/#fingerprinting-linkability

@npdoty, @dontcallmedom PTAL

### DIFF
--- a/MediaRecorder.bs
+++ b/MediaRecorder.bs
@@ -200,7 +200,7 @@ interface MediaRecorder : EventTarget {
     from this {{MediaRecorder}}, throw a {{SecurityError}} {{DOMException}} and
     abort these steps.</li>
 
-    <li>Set {{state}} to {{recording}}, and run the following steps in parallel:
+    <li>Set {{state}} to {{recording}}, and run the following steps <a>in parallel</a>:
     <ol>
       <li>Start gathering the data into a {{Blob}} <var>blob</var> and queue a
       task, using the DOM manipulation task source, to <a>fire an event</a>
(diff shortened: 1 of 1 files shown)


npdoty:
Those are a very good set of documents to start from when working on security/privacy considerations. The `/TR/` version of the fingerprinting guidance is out of date, though, we are about to publish a new version and the editor's draft may be useful in terms of identifying different severity factors and mitigations:
https://w3c.github.io/fingerprinting-guidance/
Availability (https://w3c.github.io/fingerprinting-guidance/#identifying-fingerprinting-surface-and-evaluating-severity) is a factor that might be worth considering in this section of the specification: can any embedded iframe or third-party JavaScript use these features without user interaction? (In #142 I explicitly described this as drive-by available fingerprinting surface, but I'm not sure if that's still the case.)

There are a few mitigations described at the end of the new section that use "should" language, but I'm not clear on whether that's intended to be normative or not. (It seems likely that it's not.) Are there any normative mitigations that could be described, so that sites can understand that the same protections apply in all user agents?

I wasn't clear on what the suggested mitigations were for "default values". Could you explain that further?

The first example, which immediately follows the privacy and security considerations section, is an example that explicitly shows how to enumerate the UA's supported types, which seems like an explicitly bad practice and isn't showing an example driven by functionality. Limiting the number of type supported checks might be a good mitigation to recommend to those UAs that are going to support a diversity of types.

jan-ivar (approved):
LGTM.

pes10k:
This doesn't seem to address @npdoty 's concerns above, specifically:

> I wasn't clear on what the suggested mitigations were for "default values". Could you explain that further?
> 
> The first example, which immediately follows the privacy and security considerations section, is an example that explicitly shows how to enumerate the UA's supported types, which seems like an explicitly bad practice and isn't showing an example driven by functionality. Limiting the number of type supported checks might be a good mitigation to recommend to those UAs that are going to support a diversity of types.

Is this intentional?  Does the WG disagree with the assessment / suggestion?

dontcallmedom:
@pes10k I had raised it as a separate issue #196 - I don't think the WG has managed to form an opinion on it yet, but I wanted to make sure the existing progress was merged in

pes10k:
Ah gotcha, thanks for clarifying @dontcallmedom.  That sounds good to me!

Issue #169, opened 2019-06-12
What are the default bitrate attribute values?


jan-ivar:
The [videoBitsPerSecond](https://w3c.github.io/mediacapture-record/MediaRecorder.html#dom-mediarecorder-videobitspersecond) read-only attribute says *"The value of the Video encoding target bit rate that was passed to the Platform (potentially truncated, rounded, etc), or the calculated one if the user has specified bitsPerSecond."*

This does not say what to return If neither `videoBitsPerSecond` nor `bitsPerSecond` were passed. Ditto for [audioBitsPerSecond](https://w3c.github.io/mediacapture-record/MediaRecorder.html#dom-mediarecorder-audiobitspersecond).

Possible solutions:
 1. return `0`
 2. Make `audioBitsPerSecond` and `videoBitsPerSecond` attributes nullable.

alvestrand:
My reading is that the encoder has to specify some target bitrate in order to operate at all, and this makes a means of reading what the implementation chose for you.

If it's not set, "return an UA-chosen value" seems correct to me. (Returning zero would indicate that you aren't going to produce anything unless a bitrate is specified, which seems silly.)

jan-ivar:
@alvestrand Chrome returns `0`. From console:
```
> new MediaRecorder(new MediaStream()).audioBitsPerSecond
0

> new MediaRecorder(new MediaStream()).videoBitsPerSecond
0
```

Pehrsons:
@yellowdoge your thoughts?

jan-ivar:
I'd go with @alvestrand here.

PR #171, opened 2019-06-20
Clarify MediaRecorder's mimeType handling

### LINKED ISSUE #170: Clarification needed for mime type handling
Let's start by looking at the `MediaRecorderOptions` `mimeType` attribute, where the spec says:
> The container and codec format(s) [RFC2046] for the recording, which may include any parameters that are defined for the format. If the UA does not support the format or any of the parameters specified, it MUST throw a `NotSupportedError` `DOMException`. If this paramater is not specified, the UA will use a platform-specific default format. The container format, whether passed in to the constructor or defaulted, will be used as the value of the `mimeType` attribute.

This makes it sound like the `MediaRecorder`, upon running the constructor, will lock its `mimeType` attribute to either the passed in value, or a defaulted value (if invalid it throws, so all these are valid and supported).

Then if we look at the `MediaRecorder` `mimeType` attribute, the spec says:
> The MIME type [RFC2046] that has been selected as the container for recording. This entry includes all the parameters to the base mimeType. The UA should be able to play back any of the MIME types it supports for recording. For example, it should be able to display a video recording in the HTML <video> tag. The default value for this property is platform-specific. 

This also sounds like it's read-only and never changes, but it seems to go against the language for the `MediaRecorderOptions` `mimeType` attribute in that it says "This entry includes all the parameters to the base mimeType". I take it "all the parameters" (definition?) are meant to be added to whatever was passed in with `MediaRecorderOptions`?
The contradiction lies in that the language for `MediaRecorderOptions.mimeType` says "The container format, whether passed in to the constructor or defaulted, will be used as the value of the `mimeType` attribute".

Clarifying this would be Part 1 of this issue.

Part 2 is minor, but "all the parameters" in the paragraph above would be good to clarify.
- Is it a MUST to define all codecs to be used before tracks are known? Are there other parameters that MUST be listed explicitly? "all" is very broad.

Part 3 is to clarify what we are allowed to encode once the `mimeType` has been locked onto.
- Can an audio mime type (e.g., `audio/webm;codecs=opus`) encode video?
- Can a video mime type (e.g., `video/webm;codecs=vp8`) encode audio?
- Can an audio/video mime type (e.g., `video/webm;codecs=opus,vp8`) encode only audio?
- Can an audio mime type declare a video codec and encode audio and video?

To gather some information on existing behavior I wrote a fiddle to test this, https://jsfiddle.net/pehrsons/h4rwbn1e/.
It gives per the below,

Legend:
- `ctor-mime` is the recorder's `mimeType` attribute just after running the MediaRecorder constructor
- `recording-mime` is the recorder's `mimeType` attribute just after the first `dataavailable` event (timeslice is provided)
- `blob-mime` is the `type` attribute of the `Blob` in the first `dataavailable` event
- `recorded tracks` is a list of the tracks (and their kinds) that were the result of decoding the recording, i.e., to see if the `mimeType` attribute was lying

Firefox (though note that this is the behavior I'm intending to make spec compliant):
```
[MediaRecorder "" [audio]]: ctor-mime: ""; recording-mime: "audio/ogg"; blob-mime: "audio/ogg"; recorded tracks: [audio]
[MediaRecorder "" [video]]: ctor-mime: ""; recording-mime: "video/webm"; blob-mime: "video/webm"; recorded tracks: [video]
[MediaRecorder "" [audio,video]]: ctor-mime: ""; recording-mime: "video/webm"; blob-mime: "video/webm"; recorded tracks: [audio,video]
[MediaRecorder "audio/webm;codecs=opus" [audio]]: ctor-mime: "audio/webm;codecs=opus"; recording-mime: "audio/ogg"; blob-mime: "audio/ogg"; recorded tracks: [audio]
[MediaRecorder "audio/webm;codecs=opus" [video]]: ctor-mime: "audio/webm;codecs=opus"; recording-mime: "video/webm"; blob-mime: "video/webm"; recorded tracks: [video]
[MediaRecorder "audio/webm;codecs=opus" [audio,video]]: ctor-mime: "audio/webm;codecs=opus"; recording-mime: "video/webm"; blob-mime: "video/webm"; recorded tracks: [audio,video]
[MediaRecorder "video/webm;codecs=vp8" [audio]]: ctor-mime: "video/webm;codecs=vp8"; recording-mime: "audio/ogg"; blob-mime: "audio/ogg"; recorded tracks: [audio]
[MediaRecorder "video/webm;codecs=vp8" [video]]: ctor-mime: "video/webm;codecs=vp8"; recording-mime: "video/webm"; blob-mime: "video/webm"; recorded tracks: [video]
[MediaRecorder "video/webm;codecs=vp8" [audio,video]]: ctor-mime: "video/webm;codecs=vp8"; recording-mime: "video/webm"; blob-mime: "video/webm"; recorded tracks: [audio,video]
[MediaRecorder "video/webm;codecs=opus,vp8" [audio]]: ctor-mime: "video/webm;codecs=opus,vp8"; recording-mime: "audio/ogg"; blob-mime: "audio/ogg"; recorded tracks: [audio]
[MediaRecorder "video/webm;codecs=opus,vp8" [video]]: ctor-mime: "video/webm;codecs=opus,vp8"; recording-mime: "video/webm"; blob-mime: "video/webm"; recorded tracks: [video]
[MediaRecorder "video/webm;codecs=opus,vp8" [audio,video]]: ctor-mime: "video/webm;codecs=opus,vp8"; recording-mime: "video/webm"; blob-mime: "video/webm"; recorded tracks: [audio,video]
[MediaRecorder "audio/webm;codecs=opus,vp8" [audio]]: Exception; NotSupportedError: Operation is not supported
[MediaRecorder "audio/webm;codecs=opus,vp8" [video]]: Exception; NotSupportedError: Operation is not supported
[MediaRecorder "audio/webm;codecs=opus,vp8" [audio,video]]: Exception; NotSupportedError: Operation is not supported
```

Chrome:
```
[MediaRecorder "" [audio]]: ctor-mime: "audio/webm;codecs=opus"; recording-mime: "audio/webm;codecs=opus"; blob-mime: "audio/webm;codecs=opus"; recorded tracks: [audio]
[MediaRecorder "" [video]]: ctor-mime: "video/webm;codecs=vp8"; recording-mime: "video/webm;codecs=vp8"; blob-mime: "video/webm;codecs=vp8"; recorded tracks: [video]
[MediaRecorder "" [audio,video]]: ctor-mime: "video/webm;codecs=vp8,opus"; recording-mime: "video/webm;codecs=vp8,opus"; blob-mime: "video/webm;codecs=vp8,opus"; recorded tracks: [audio,video]
[MediaRecorder "audio/webm;codecs=opus" [audio]]: ctor-mime: "audio/webm;codecs=opus"; recording-mime: "audio/webm;codecs=opus"; blob-mime: "audio/webm;codecs=opus"; recorded tracks: [audio]
[MediaRecorder "audio/webm;codecs=opus" [video]]: ctor-mime: "audio/webm;codecs=opus"; recording-mime: "audio/webm;codecs=opus"; blob-mime: "audio/webm;codecs=opus"; recorded tracks: [video]
[MediaRecorder "audio/webm;codecs=opus" [audio,video]]: ctor-mime: "audio/webm;codecs=opus"; recording-mime: "audio/webm;codecs=opus"; blob-mime: "audio/webm;codecs=opus"; recorded tracks: [audio,video]
[MediaRecorder "video/webm;codecs=vp8" [audio]]: ctor-mime: "video/webm;codecs=vp8"; recording-mime: "video/webm;codecs=vp8"; blob-mime: "video/webm;codecs=vp8"; recorded tracks: [audio]
[MediaRecorder "video/webm;codecs=vp8" [video]]: ctor-mime: "video/webm;codecs=vp8"; recording-mime: "video/webm;codecs=vp8"; blob-mime: "video/webm;codecs=vp8"; recorded tracks: [video]
[MediaRecorder "video/webm;codecs=vp8" [audio,video]]: ctor-mime: "video/webm;codecs=vp8"; recording-mime: "video/webm;codecs=vp8"; blob-mime: "video/webm;codecs=vp8"; recorded tracks: [audio,video]
[MediaRecorder "video/webm;codecs=opus,vp8" [audio]]: ctor-mime: "video/webm;codecs=opus,vp8"; recording-mime: "video/webm;codecs=opus,vp8"; blob-mime: "video/webm;codecs=opus,vp8"; recorded tracks: [audio]
[MediaRecorder "video/webm;codecs=opus,vp8" [video]]: ctor-mime: "video/webm;codecs=opus,vp8"; recording-mime: "video/webm;codecs=opus,vp8"; blob-mime: "video/webm;codecs=opus,vp8"; recorded tracks: [video]
[MediaRecorder "video/webm;codecs=opus,vp8" [audio,video]]: ctor-mime: "video/webm;codecs=opus,vp8"; recording-mime: "video/webm;codecs=opus,vp8"; blob-mime: "video/webm;codecs=opus,vp8"; recorded tracks: [audio,video]
[MediaRecorder "audio/webm;codecs=opus,vp8" [audio]]: Exception; NotSupportedError: Failed to construct 'MediaRecorder': Failed to initialize native MediaRecorder the type provided (audio/webm;codecs=opus,vp8) is not supported.
[MediaRecorder "audio/webm;codecs=opus,vp8" [video]]: Exception; NotSupportedError: Failed to construct 'MediaRecorder': Failed to initialize native MediaRecorder the type provided (audio/webm;codecs=opus,vp8) is not supported.
[MediaRecorder "audio/webm;codecs=opus,vp8" [audio,video]]: Exception; NotSupportedError: Failed to construct 'MediaRecorder': Failed to initialize native MediaRecorder the type provided (audio/webm;codecs=opus,vp8) is not supported.
```

Safari: Not implemented

Given this, I think we need Part 4: Should the `MediaRecorder` `mimeType` attribute really be read-only and never change? While I do agree this makes for a simpler, less confusing, API, no implementation works this way today.

---

To recap:

1: Can the `MediaRecorder` constructor modify the `MediaRecorderOptions` `mimeType` attribute as it sets its own `mimeType` attribute?
2: Define "all the parameters" for a mime type, or rephrase. In particular, is it a MUST to include a final list of codecs?
3: What restrictions on tracks does the selected `mimeType` impose?
4: Can the `MediaRecorder` `mimeType` attribute change during the `MediaRecorder`s lifetime (to maintain behavior of current implementations)?

Pehrsons:
Fixes #170.

This lets the mimeType passed to a recorder on construction be
constraining, to force a recorder to use a certain container or codecs.
It also updates a recorder's mimeType while it's recording, so the
application can be aware of the decisions the UA makes for it.

### DIFF
--- a/MediaRecorder.bs
+++ b/MediaRecorder.bs
@@ -111,8 +111,7 @@ interface MediaRecorder : EventTarget {
         "False">&#10008;</span></td>
         <td class="prmOptTrue"><span role="img" aria-label=
         "True">&#10004;</span></td>
-        <td class="prmDesc">A dictionary of options to for the UA instructing
-        how the recording will take part.</td>
+        <td class="prmDesc">A dictionary of recording options.</td>
       </tr>
     </tbody>
   </table>
@@ -120,20 +119,15 @@ interface MediaRecorder : EventTarget {
   <dd algorithm="to construct a MediaRecorder">When the {{MediaRecorder()}}
   constructor is invoked, the User Agent MUST run the following steps:
   <ol>
-    <li>Let |recorder| be the newly constructed {{MediaRecorder}}.</li>
+    <li>Let |recorder| be a newly constructed {{MediaRecorder}} object.</li>
     <li>Initialize |recorder|'s {{MediaRecorder/stream}} attribute to the
     constructor's first argument.</li>
     <li>Each {{MediaRecorder}} has an internal <dfn>constrainedMimeType</dfn>
     property that is initially set to the empty string.</li>
-    <li>If the constructor's second argument |options| is present, run the
-    following steps:
-    <ol>
-      <li>If |options|' {{MediaRecorderOptions/mimeType}} member is present, set
-      the [=constrainedMimeType=] internal property to the value of |options|'
-      {{MediaRecorderOptions/mimeType}} member.</li>
-    </ol>
-    </li>
-    <li>If invoking {{isTypeSupported()}} with the [=constrainedMimeType=]
+    <li>If the {{MediaRecorderOptions/mimeType}} member is present in the
+    constructor's second argument, set the [=constrainedMimeType=] internal
+    property to the value of {{MediaRecorderOptions/mimeType}}.</li>
+    <li>If invoking [$is type supported$] with the [=constrainedMimeType=]
     internal property as its argument returns false, throw a
     {{NotSupportedError}} {{DOMException}} and abort these steps.</li>
     <li>Initialize |recorder|'s {{MediaRecorder/mimeType}} attribute to the
@@ -227,24 +221,33 @@ interface MediaRecorder : EventTarget {
     [=track set=].</li>
 
     <li>If the [=constrainedMimeType=] internal property specifies a media type,
-    container, or codec such that the User Agent cannot record a track in
-    |tracks| with it, throw a {{NotSupportedError}} {{DOMException}} and abort
-    these steps.</li>
+    container, or codec, then run the following sub steps:
+    <ol>
+      <li>Constrain the configuration of the MediaRecorder to the media type,
+      container, and codec specified in [=constrainedMimeType=].</li>
+      <li>For each track in |tracks|, if the User Agent cannot record the track
+      using the current configuration, then throw a {{NotSupportedError}}
+      {{DOMException}} and abort all steps.</li>
+    </ol>
+    </li>
 
-    <li>Set {{mimeType}} to the [=constrainedMimeType=] internal property,
-    extended by the media type, subtype and codecs parameter reflecting the
-    configuration used by the MediaRecorder to record all tracks in |tracks|.
-    The media type, subtype and any parameters in the [=constrainedMimeType=]
-    internal property, MUST NOT be changed. {{mimeType}} MAY be extended with
-    the profiles parameter [[RFC6381]] or further codec-specific parameters to
-    avoid ambiguity.</li>
+    <li>Let |extendedMimeType| be the value of the [=constrainedMimeType=]
+    internal property.</li>
+
+    <li>Modify |extendedMimeType| by adding media type, subtype and codecs
+    parameter reflecting the configuration used by the MediaRecorder to record
+    all tracks in |tracks|, if not already present. This MAY include the
+    profiles parameter [[RFC6381]] or further codec-specific parameters.</li>
+
+    <li>Set {{mimeType}} to |extendedMimeType|.</li>
 
     <li>Set {{state}} to {{recording}}, and run the following steps in parallel:
     <ol>
-      <li>Start recording all tracks in |tracks| and gather the data into a
-      {{Blob}} <var>blob</var> and queue a task, using the DOM manipulation task
-      source, to <a>fire an event</a> named <a>start</a> at <var>target</var>,
-      then return <code>undefined</code>.</li>
+      <li>Start recording all tracks in |tracks| using the {{MediaRecorder}}'s
+      current configuration, and gather the data into a {{Blob}} <var>blob</var>
+      and queue a task, using the DOM manipulation task source, to
+      <a>fire an event</a> named <a>start</a> at <var>target</var>, then return
+      <code>undefined</code>.</li>
 
       <li>If at any point the {{stream}}'s <a>isolation properties</a> change so
       that {{MediaRecorder}} is no longer allowed access to it, the UA MUST
@@ -458,26 +461,8 @@ interface MediaRecorder : EventTarget {
   {{MediaRecorder}} implementation is capable of recording {{Blob}} objects for
   the specified MIME type. Recording may still fail if sufficient resources are
   not available to support the concrete media encoding. When this method is
-  invoked, the User Agent must run the following steps:
-  <ol class="method-algorithm">
-    <li>If <code>type</code> is an empty string, then return true
-    (note that this case is essentially equivalent to leaving up to
-    the UA the choice of both container and codecs).</li>
-    <li>If <code>type</code> does not contain a valid MIME type
-    string, then return false.</li>
-    <li>If <code>type</code> contains a media type or media subtype
-    that the MediaRecorder does not support, then return false.</li>
-    <li>If <code>type</code> contains a media container that the
-    MediaRecorder does not support, then return false.</li>
-    <li>If <code>type</code> contains more than one audio codec, or more than
-    one video codec, then return false.</li>
-    <li>If <code>type</code> contains a codec that the MediaRecorder
-    does not support, then return false.</li>
-    <li>If the MediaRecorder does not support the specified
-    combination of media type/subtype, codecs and container then
-    return false.</li>
-    <li>Return true.</li>
-  </ol>
+  invoked, the User Agent must return the result of [$is type supported$],
+  given the method's first argument.
   <table class="parameters">
     <tbody>
       <tr>
@@ -500,6 +485,28 @@ interface MediaRecorder : EventTarget {
       </tr>
     </tbody>
   </table>
+  </dd>
+  <dd algorithm="is type supported">The <dfn abstract-op>is type supported</dfn>
+  algorithm consists of the following steps.
+  <ol>
+    <li>Let |type| be the algorithm's argument.</li>
+    <li>If |type| is an empty string, then return true (note that this case is
+    essentially equivalent to leaving up to the UA the choice of both container
+    and codecs).</li>
+    <li>If |type| does not contain a valid MIME type string, then return false.
+    </li>
+    <li>If |type| contains a media type or media subtype that the MediaRecorder
+    does not support, then return false.</li>
+    <li>If |type| contains a media container that the MediaRecorder does not
+    support, then return false.</li>
+    <li>If |type| contains more than one audio codec, or more than one video
+    codec, then return false.</li>
+    <li>If |type| contains a codec that the MediaRecorder does not support, then
+    return false.</li>
+    <li>If the MediaRecorder does not support the specified combination of media
+    type/subtype, codecs and container then return false.</li>
+    <li>Return true.</li>
+  </ol>
   </dd>
 </dl>
 
--- a/MediaRecorder.bs
+++ b/MediaRecorder.bs
@@ -112,13 +112,36 @@ interface MediaRecorder : EventTarget {
         <td class="prmOptTrue"><span role="img" aria-label=
         "True">&#10004;</span></td>
         <td class="prmDesc">A dictionary of options to for the UA instructing
-        how the recording will take part.
-        {{MediaRecorderOptions/mimeType|options.mimeType}}, if present, will
-        become the value of {{MediaRecorder/mimeType}} attribute.</td>
+        how the recording will take part.</td>
       </tr>
     </tbody>
   </table>
   </dd>
+  <dd algorithm="to construct a MediaRecorder">When the {{MediaRecorder()}}
+  constructor is invoked, the User Agent MUST run the following steps:
+  <ol>
+    <li>Let |recorder| be the newly constructed {{MediaRecorder}}.</li>
+    <li>Initialize |recorder|'s {{MediaRecorder/stream}} attribute to the
+    constructor's first argument.</li>
+    <li>Each {{MediaRecorder}} has an internal <dfn>constrainedMimeType</dfn>
+    property that is initially set to the empty string.</li>
+    <li>If the constructor's second argument |options| is present, run the
+    following steps:
+    <ol>
+      <li>If |options|' {{MediaRecorderOptions/mimeType}} member is present, set
+      the [=constrainedMimeType=] internal property to the value of |options|'
+      {{MediaRecorderOptions/mimeType}} member.</li>
+    </ol>
+    </li>
+    <li>If invoking {{isTypeSupported()}} with the [=constrainedMimeType=]
+    internal property as its argument returns false, throw a
+    {{NotSupportedError}} {{DOMException}} and abort these steps.</li>
+    <li>Initialize |recorder|'s {{MediaRecorder/mimeType}} attribute to the
+    [=constrainedMimeType=] internal property.</li>
+    <li>Initialize |recorder|'s {{MediaRecorder/state}} attribute to
+    {{inactive}}.</li>
+    <li>Return |recorder|.</li>
+  </ol></dd>
 </dl>
 
 ## Attributes ## {#mediarecorder-attributes}
@@ -128,12 +151,10 @@ interface MediaRecorder : EventTarget {
   <dd>The {{MediaStream}} [[!GETUSERMEDIA]] to be recorded.</dd>
 
   <dt><dfn attribute for="MediaRecorder"><code>mimeType</code></dfn></dt>
-  <dd>The MIME type [[!RFC2046]] that has been selected as the container
-  for recording. This entry includes all the parameters to the base
-  <code>mimeType</code>. The UA should be able to play back any of the
-  MIME types it supports for recording. For example, it should be able
-  to display a video recording in the HTML &lt;video&gt; tag. The
-  default value for this property is platform-specific.
+  <dd>The MIME type [[!RFC2046]] used by the {{MediaRecorder}} object.
+  The User Agent SHOULD be able to play back any of the MIME types it supports
+  for recording. For example, it should be able to display a video recording in
+  the HTML &lt;video&gt; tag.
 
   <div class="note"> {{MediaRecorder/mimeType}} specifies the media type and
   container format for the recording via a type/subtype combination, with the
@@ -142,8 +163,7 @@ interface MediaRecorder : EventTarget {
   </div></dd>
 
   <dt><dfn attribute for="MediaRecorder"><code>state</code></dfn></dt>
-  <dd>The current state of the {{MediaRecorder}} object. When the
-  {{MediaRecorder}} is created, the UA MUST set this attribute to {{inactive}}.
+  <dd>The current state of the {{MediaRecorder}} object.
   </dd>
 
   <dt><dfn attribute for="MediaRecorder"><code>onstart</code></dfn></dt>
@@ -185,7 +205,7 @@ interface MediaRecorder : EventTarget {
     and fire events asynchronously.
   </div>
   <dt><dfn method for="MediaRecorder"><code>start(optional unsigned long timeslice)</code></dfn></dt>
-  <dd>
+  <dd algorithm="to start a MediaRecorder">
   When a {{MediaRecorder}} object&#8217;s {{start()}} method is invoked, the UA
   MUST run the following steps:
   <ol>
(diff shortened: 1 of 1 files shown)


Pehrsons:
@jan-ivar PTAL

MediaRecorder.bs
```diff
@@ -112,13 +112,36 @@ interface MediaRecorder : EventTarget {
         <td class="prmOptTrue"><span role="img" aria-label=
         "True">&#10004;</span></td>
         <td class="prmDesc">A dictionary of options to for the UA instructing
-        how the recording will take part.
```
jan-ivar, line 115: Can we fix this sentence while we're here? How about *"A dictionary of recording options."*

MediaRecorder.bs
```diff
@@ -112,13 +112,36 @@ interface MediaRecorder : EventTarget {
...
-        become the value of {{MediaRecorder/mimeType}} attribute.</td>
+        how the recording will take part.</td>
       </tr>
     </tbody>
   </table>
   </dd>
+  <dd algorithm="to construct a MediaRecorder">When the {{MediaRecorder()}}
+  constructor is invoked, the User Agent MUST run the following steps:
+  <ol>
+    <li>Let |recorder| be the newly constructed {{MediaRecorder}}.</li>
```
jan-ivar: Let |recorder| be a newly constructed {{MediaRecorder}} object.

MediaRecorder.bs
```diff
@@ -112,13 +112,36 @@ interface MediaRecorder : EventTarget {
...
+    <li>Initialize |recorder|'s {{MediaRecorder/stream}} attribute to the
+    constructor's first argument.</li>
+    <li>Each {{MediaRecorder}} has an internal <dfn>constrainedMimeType</dfn>
+    property that is initially set to the empty string.</li>
```
jan-ivar: I prefer instructive to descriptive in algorithms. Also, this spec doesn't appear to have an established pattern for internal slots. Did you model this on a specific spec?

If not, we could follow WebRTC:
```
Let |recorder| have a <dfn>\[[ConstrainedMimeType]]</dfn> internal slot initialized to
the empty string.
```
Happy to follow something else as well, as long as there's a precedent.

MediaRecorder.bs
```diff
@@ -112,13 +112,36 @@ interface MediaRecorder : EventTarget {
...
+    <li>If the constructor's second argument |options| is present, run the
+    following steps:
```
jan-ivar: Optional dictionaries are [always present](https://heycam.github.io/webidl/#dfn-optional-argument-default-value) as method arguments:
> If the type of an argument is a dictionary type or a union type that has a dictionary as one of its flattened member types, and that dictionary type and its ancestors have no required members, and the argument is either the final argument or is followed only by optional arguments, then the argument must be specified as optional. Such arguments are **always considered to have a default value of an empty dictionary**, unless otherwise specified.

So we don't need this check.

MediaRecorder.bs
```diff
@@ -112,13 +112,36 @@ interface MediaRecorder : EventTarget {
...
+    <ol>
+      <li>If |options|' {{MediaRecorderOptions/mimeType}} member is present, set
+      the [=constrainedMimeType=] internal property to the value of |options|'
+      {{MediaRecorderOptions/mimeType}} member.</li>
+    </ol>
+    </li>
+    <li>If invoking {{isTypeSupported()}} with the [=constrainedMimeType=]
```
jan-ivar: [isTypeSupported](https://w3c.github.io/mediacapture-record/#dom-mediarecorder-istypesupported) is a webidl method, and we generally don't want to invoke these from algorithms, as it may end up running unnecessary binding code.

I'd move its steps into a new **is type supported** algorithm, and then for *isTypeSupported* say *"When this method is invoked, the User Agent must return the result of [is type supported](https://w3c.github.io/mediacapture-record/#dom-mediarecorder-istypesupported), given the method's first argument."*

MediaRecorder.bs
```diff
@@ -128,12 +151,10 @@ interface MediaRecorder : EventTarget {
   <dd>The {{MediaStream}} [[!GETUSERMEDIA]] to be recorded.</dd>
 
   <dt><dfn attribute for="MediaRecorder"><code>mimeType</code></dfn></dt>
-  <dd>The MIME type [[!RFC2046]] that has been selected as the container
-  for recording. This entry includes all the parameters to the base
-  <code>mimeType</code>. The UA should be able to play back any of the
-  MIME types it supports for recording. For example, it should be able
-  to display a video recording in the HTML &lt;video&gt; tag. The
-  default value for this property is platform-specific.
```
jan-ivar, line 136: Why lose the last sentence? Isn't it still true?

MediaRecorder.bs
```diff
@@ -112,13 +112,36 @@ interface MediaRecorder : EventTarget {
...
+    </ol>
+    </li>
+    <li>If invoking {{isTypeSupported()}} with the [=constrainedMimeType=]
+    internal property as its argument returns false, throw a
+    {{NotSupportedError}} {{DOMException}} and abort these steps.</li>
+    <li>Initialize |recorder|'s {{MediaRecorder/mimeType}} attribute to the
+    [=constrainedMimeType=] internal property.</li>
+    <li>Initialize |recorder|'s {{MediaRecorder/state}} attribute to
+    {{inactive}}.</li>
+    <li>Return |recorder|.</li>
```
jan-ivar, line 139: Now that we have an explicit constructor algorithm, it must be exhaustive. You've done {{state}}, but we're still missing audioBitsPerSecond, videoBitsPerSecond and bitsPerSecond.

MediaRecorder.bs
```diff
@@ -196,23 +216,44 @@ interface MediaRecorder : EventTarget {
     <li>If {{state}} is not {{inactive}}, throw an {{InvalidStateError}}
     {{DOMException}} and abort these steps.</li>
 
-    <li>If the {{MediaRecorder/stream}}'s <a>isolation properties</a> disallow access
+    <li>If the {{stream}}'s <a>isolation properties</a> disallow access
     from this {{MediaRecorder}}, throw a {{SecurityError}} {{DOMException}} and
     abort these steps.</li>
 
+    <li>If the {{stream}} is {{inactive}}, throw a {{NotSupportedError}}
+    {{DOMException}} and abort these steps.</li>
```
jan-ivar, line 227: This is new. While Chrome appears to throw an error here ("UnknownError"), Firefox [does not](https://jsfiddle.net/jib1/gmb5andt/). In fact, firefox lets me add tracks to the stream right after `start()` and it still records them. Is your plan to fix Firefox?

MediaRecorder.bs
```diff
@@ -196,23 +216,44 @@ interface MediaRecorder : EventTarget {
...
+
+    <li>Let |tracks| be the set of {{live}} tracks in the {{stream}}'s
+    [=track set=].</li>
+
+    <li>If the [=constrainedMimeType=] internal property specifies a media type,
+    container, or codec such that the User Agent cannot record a track in
```
jan-ivar: This seems to imply that recording will be constrained by [=constrainedMimeType=], but doesn't actually say so. How about:

"If the [=constrainedMimeType=] internal property specifies a media type, container, or codec, then run these sub steps:
 1. Constrain the configuration of the MediaRecorder to the media type, container, and codec specified in [=constrainedMimeType=].
 2. for each track in *tracks*, if the User Agent cannot cannot record the track using the current configuration, then throw a {{NotSupportedError}} {{DOMException}} and abort all steps.
" ?

MediaRecorder.bs
```diff
@@ -196,23 +216,44 @@ interface MediaRecorder : EventTarget {
...
+    |tracks| with it, throw a {{NotSupportedError}} {{DOMException}} and abort
+    these steps.</li>
+
+    <li>Set {{mimeType}} to the [=constrainedMimeType=] internal property,
+    extended by the media type, subtype and codecs parameter reflecting the
+    configuration used by the MediaRecorder to record all tracks in |tracks|.
+    The media type, subtype and any parameters in the [=constrainedMimeType=]
+    internal property, MUST NOT be changed. {{mimeType}} MAY be extended with
+    the profiles parameter [[RFC6381]] or further codec-specific parameters to
+    avoid ambiguity.</li>
```
jan-ivar: Instead of saying constrainedMimeType MUST NOT be changed, maybe split this into steps to make it clear? E.g.

 * Let |extendedMimeType| be the value of [=constrainedMimeType=]
 * Modify |extendedMimeType| by adding media type, subtype and codecs parameter reflecting the configuration used by the MediaRecorder to record all tracks in |tracks|, if not already present.
 * Set {{mimeType}} to |extendedMimeType|

MediaRecorder.bs
```diff
@@ -196,23 +216,44 @@ interface MediaRecorder : EventTarget {
...
-      task, using the DOM manipulation task source, to <a>fire an event</a>
-      named <a>start</a> at <var>target</var>, then return <code>undefined</code>.
-      </li>
-
-      <li>If at any point the {{MediaRecorder/stream}}'s <a>isolation properties</a>
-      change so that {{MediaRecorder}} is no longer allowed access to it, the
-      UA MUST immediately stop gathering data, discard any data that it has
-      gathered, and queue a task, using the DOM manipulation task source, that
-      runs the following steps:
+      <li>Start recording all tracks in |tracks| and gather the data into a
```
jan-ivar: Maybe "start recording all tracks in |tracks| using the MediaRecorder's current configuration, and gather..."?

MediaRecorder.bs
```diff
@@ -196,23 +216,44 @@ interface MediaRecorder : EventTarget {
...
+      then return <code>undefined</code>.</li>
+
+      <li>If at any point the {{stream}}'s <a>isolation properties</a> change so
+      that {{MediaRecorder}} is no longer allowed access to it, the UA MUST
+      immediately stop gathering data, discard any data that it has gathered,
+      and queue a task, using the DOM manipulation task source, that runs the
+      following steps:
       <ol>
+        <li>Set {{mimeType}} to the value of the [=constrainedMimeType=]
+        internal property.</li>
```
jan-ivar: Do browsers currently do this?

MediaRecorder.bs
```diff
@@ -222,10 +263,13 @@ interface MediaRecorder : EventTarget {
...
-      <li>If at any point, a track is added to or removed from the {{MediaRecorder/stream}}'s
-      track set, the UA MUST immediately stop gathering data, discard any data that it has gathered,
-      and queue a task, using the DOM manipulation task source, that runs the following steps:
+      <li>If at any point, a track is added to or removed from the {{stream}}'s
+      [=track set=], the UA MUST immediately stop gathering data, discard any
+      data that it has gathered, and queue a task, using the DOM manipulation
+      task source, that runs the following steps:
       <ol>
+        <li>Set {{mimeType}} to the value of the [=constrainedMimeType=]
+        internal property.</li>
```
jan-ivar: Ditto.

jan-ivar (changes requested):
Thanks! This adds more specificity to algorithms, which is great! But also means I have lots of nits and questions.

MediaRecorder.bs
Pehrsons: I followed this precedent from @martinthomson 
https://w3c.github.io/mediacapture-fromelement/#dfn-framecapturerequested

MediaRecorder.bs
Pehrsons: No, the constructor algorithm initializes it to MediaRecorderOptions/mimeType (explicit).
Then defaults can come into effect in `start()` but that's handled by its algorithm -- which says to reflect the actual configuration used for recording in the `mimeType` attribute, leaving to the UA to choose it freely, unless constrained by the ctor. In any case there is no default value to the `mimeType` attribute.

MediaRecorder.bs
Pehrsons: So that's pending the resolution of #169. Should I put up a separate PR for that issue, or even fix it in this PR?

MediaRecorder.bs
Pehrsons: Indeed. I will fix Firefox in [bug 1512175](https://bugzilla.mozilla.org/show_bug.cgi?id=1512175).

MediaRecorder.bs
Pehrsons: Firefox does sort of, it sets `mimeType` to "" before firing "stop", https://searchfox.org/mozilla-central/rev/0b7007a23bc16c857f894140e12f307bfeef2fdd/dom/media/MediaRecorder.cpp#396-397

Chrome doesn't. It selects `mimeType` in the constructor and doesn't change it even if using a different configuration to encode tracks.

Here's a modified fiddle showing this behavior, https://jsfiddle.net/pehrsons/1Lznhjyg/

Exemplified, Firefox:
```
[MediaRecorder "" [audio,video]]: ctor-mime: ""; recording-mime: "video/webm"; stopped-mime: ""; "blob-mime: "video/webm"; recorded tracks: [audio,video]
[MediaRecorder "audio/webm;codecs=opus" [video]]: ctor-mime: "audio/webm;codecs=opus"; recording-mime: "video/webm"; stopped-mime: ""; "blob-mime: "video/webm"; recorded tracks: [video]
[MediaRecorder "video/webm;codecs=opus,vp8" [audio,video]]: ctor-mime: "video/webm;codecs=opus,vp8"; recording-mime: "video/webm"; stopped-mime: ""; "blob-mime: "video/webm"; recorded tracks: [audio,video]
```

Chrome:
```
[MediaRecorder "" [audio,video]]: ctor-mime: ""; recording-mime: ""; stopped-mime: ""; "blob-mime: ""; recorded tracks: [audio,video]
[MediaRecorder "audio/webm;codecs=opus" [video]]: ctor-mime: "audio/webm;codecs=opus"; recording-mime: "audio/webm;codecs=opus"; stopped-mime: "audio/webm;codecs=opus"; "blob-mime: "audio/webm;codecs=opus"; recorded tracks: [video]
[MediaRecorder "video/webm;codecs=opus,vp8" [audio,video]]: ctor-mime: "video/webm;codecs=opus,vp8"; recording-mime: "video/webm;codecs=opus,vp8"; stopped-mime: "video/webm;codecs=opus,vp8"; "blob-mime: "video/webm;codecs=opus,vp8"; recorded tracks: [audio,video]
```

MediaRecorder.bs
Pehrsons: See above.

MediaRecorder.bs
jan-ivar: That seems like an [outlier](https://github.com/w3c/mediacapture-fromelement/commit/9364508e012d5e7cd58a203fe87ca80a6ccebb8a). My advice would be to follow [es6 terminology](https://w3c.github.io/mediacapture-main/getusermedia.html#ecmascript-6-terminology) like mediacapture-main and webrtc, which seems [well established](https://github.com/whatwg/streams/issues/178), and puts more emphasis on the difference between a webidl exposed property getter—with its binding machinery—vs an "internal slot".

MediaRecorder.bs
jan-ivar: If the idea is that the codec doesn't change, then I wouldn't expect it to change on stop either. So I'd go with Chrome here.

MediaRecorder.bs
jan-ivar: Also, the descriptive phrase style copied here was [not from an algorithm](https://w3c.github.io/mediacapture-fromelement/#methods-0), so I think we should use imperative.

MediaRecorder.bs
jan-ivar: As long as we don't forget to do it, I guess either way might work.

MediaRecorder.bs
jan-ivar: Resolved offline. Since `stop()` is not terminal, it makes some sense to revert to `constrainedMimeType` here to avoid confusion about whether the returned value is a constrained mimeType or merely a current mimeType.

jan-ivar (approved):
Thanks! I'd still prefer es6 internal slots, and we need to not forget https://github.com/w3c/mediacapture-record/issues/169 but I see no reason to block over it.

jan-ivar (approved):
👍

jan-ivar:
@armax00 PTAL

PR #172, opened 2019-06-23
Input width, height MUST be recorded and playable

guest271314:
Input MediaStreamTrack of kind "video" width and height MUST be recorded and playable.

***
<a href="https://pr-preview.s3.amazonaws.com/guest271314/mediacapture-record/pull/172.html" title="Last updated on Feb 15, 2021, 7:32 AM UTC (1ea1b20)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/mediacapture-record/172/963bfbb...guest271314:1ea1b20.html" title="Last updated on Feb 15, 2021, 7:32 AM UTC (1ea1b20)">Diff</a>

### DIFF
--- a/MediaRecorder.bs
+++ b/MediaRecorder.bs
@@ -284,6 +284,13 @@ interface MediaRecorder : EventTarget {
   (because of {{timeslice}} or {{requestData()}}), the individual Blobs need not
   be playable, but the combination of all the Blobs from a completed recording
   MUST be playable.</p>
+  
+  <p>The UA MUST record {{MediaRecorder/stream}} in such a way that the original
+  video Track(s) width and height can be retrieved at playback time. The {{Blob}}
+  containing entire recording returned when the recording finishes or smaller 
+  buffers of data returned at regular intervals containing the combination of all 
+  the original video Track(s) width and height MUST playback the input video Track(s) 
+  width and height.</p>
 
   <p> If any Track within the {{MediaStream}} is {{muted}} or not {{enabled}} at
   any time, the UA will only record black frames or silence since that is the


guest271314:
Fixes https://github.com/w3c/mediacapture-record/issues/173

Pehrsons:
It seems reasonable to expect the video to be recorded with dynamically changing resolution, so that input frames are correctly represented in the output. The main issue I see though, is if recording into a container that doesn't support changing resolution dynamically. Should that be an error? Or should it be left to the UA to take an alternative approach, like scale, crop, etc?

I don't know the container landscape well enough to say whether, or which, containers support this, but I imagine there are some that don't. Others might know better.

guest271314:
> It seems reasonable to expect the video to be recorded with dynamically changing resolution, so that input frames are correctly represented in the output. The main issue I see though, is if recording into a container that doesn't support changing resolution dynamically. Should that be an error? Or should it be left to the UA to take an alternative approach, like scale, crop, etc?
> 
> I don't know the container landscape well enough to say whether, or which, containers support this, but I imagine there are some that don't. Others might know better.

AFAIK the only container that Firefox and Chromium support is Matroska (multiple codecs supported), or Webm (Matroska with VP8, VP9, was Vorbis, now Opus). The codecs that those browsers' implement contains code which checks the width and height of input.

This is the edge case (observable at Chromium) which led to filing this issue https://bugs.chromium.org/p/chromium/issues/detail?id=972470 and this PR. Perhaps there is a bug in the code. Can you get the recording the output the width and height changes at Chromium?

```
<!DOCTYPE html>
<html>
<head>
</head>
<body>
  <video id="sink" autoplay muted controls></video>
  <script>
    (async() => {
      const sink = document.querySelector("video");
      const canvas = document.createElement("canvas");
      const imageData = [{
        width: 50,
        height: 50,
        fillStyle: "green"
      }, {
        width: 150,
        height: 150,
        fillStyle: "yellow"
      }, {
        width: 300,
        height: 300,
        fillStyle: "red"
      }];
      const segments = 60;
      const videoData = [];
      const ctx = canvas.getContext("2d");
      const canvasStream = canvas.captureStream(0);
      const [videoTrack] = canvasStream.getVideoTracks();
      const mediaStream = [canvasStream, videoTrack].find(({requestFrame:rF}) => rF);
      const recorder = new MediaRecorder(canvasStream);
      const result = new Promise(resolve =>
        recorder.ondataavailable = e => {
          sink.load();
          sink.srcObject = null;
          resolve(sink.src = URL.createObjectURL(e.data));
        });
      recorder.start();
      videoTrack.onunmute = e => console.log(e);
      videoTrack.onmute = e => console.log(e);
      videoTrack.onended = e => console.log(e);
      sink.onloadedmetadata = e => console.log(e.type, e.target.tagName, sink.videoWidth, sink.videoHeight);
      sink.onresize = e => console.log(e.type, e.target.tagName, sink.videoWidth, sink.videoHeight);
      sink.srcObject = canvasStream;

      for (const {width, height, fillStyle} of imageData) {
        // At Chromium 76 the only code that have been able to set width and height with
        // results in the incorrect value for height: 150 when width: 300
        // for last 60 frames, should be width: 300 when height: 300 
        // ({width:canvas, height:canvas});
        // Object.assign(canvas, {width, height});
        console.log(canvas.width, canvas.height, {width, height});
        // at Firefox 69 each set width and heiht of canvas
        // is reflected at resulting webm video by settin width and height
        // this should be the only setting needed and output the same result at Chromium
        canvas.width = width;
        canvas.height = height;
        await new Promise(async resolve => {
          for (let i = 0; i < segments; i++) {
            await new Promise(promise =>
              requestAnimationFrame(function draw() {
                ctx.clearRect(0, 0, width, height);
                ctx.fillStyle = fillStyle;
                ctx.fillRect(0, 0, width, height);
                mediaStream.requestFrame();
                promise();
              })
            )
          }
          resolve()
        })
      }
      recorder.stop();
      videoTrack.stop();
      videoTrack.enabled = false;
      console.log(await result, sink.srcObject, canvasStream, videoTrack);
    })();
  </script>
</body>
</html>
```
Interestingly, while experiementing with code other than the edge at Chromium the width and height is recorded when `<canvas>` `width` and `height` is changed.

If a container does not support variable width the container should not be implemented by `MediaRecorder`. 

Had not considered whether or not an error should be thrown. At first impression, variable width and height input should "just work", else the entire concept of `canvas.captureStream()` needs to be revisted and constraints should be deleted from the several revelant specifications which state they could remotely have an effect on a `MediaStreamTrack`. The error would be with the implementation by browser authors. Though if your question is essentially if an error should not be thrown then what is the point of this PR then yes, throw the error; which would require some internal code and testing to verify that the output width and height corresponds to the input width and height - which should already be implemented.

guest271314:
@Pehrsons Note, Mozilla appears to be reluctant to add support for container other than WebM and codecs other than VP8 and Opus https://bugzilla.mozilla.org/show_bug.cgi?id=1422891#c11, https://hg.mozilla.org/mozilla-central/annotate/58f1d879059c62ddeaf4b4fcf2624fb2c66d6ab9/dom/media/MediaRecorder.cpp#l1488

Pehrsons:
> @Pehrsons Note, Mozilla appears to be reluctant to add support for container other than WebM and codecs other than VP8 and Opus https://bugzilla.mozilla.org/show_bug.cgi?id=1422891#c11, https://hg.mozilla.org/mozilla-central/annotate/58f1d879059c62ddeaf4b4fcf2624fb2c66d6ab9/dom/media/MediaRecorder.cpp#l1488

I think you're overanalyzing things.

guest271314:
@Pehrsons 

> I think you're overanalyzing things.

That is a very interesting statement. How does an individual determine when a subject matter or topic has been overanalyzed? Consider the subject matter of "global warming" or "climate change", where there is no universal consensus among the parties (scientists, policy makers, think tanks, Joe and Jane Does reading the daily fish wrapper).

On the record _Angry Enuff_ by Suga Free there is an instructive line

> Look for lies! Misconduct is at an all-time high too

which can be applied to any field of endeveavor.

A trained eye and the satisfaction of the individual research scientist is required for subject matter to be concluded as overanalyzed. For some research topics it is not possible to overanalyze the data, as no conclusion has yet been reached.

- It would probably be best to overanalyze the programs and math involved in manned space flight rather than to underanalyze the mission. (https://en.wikipedia.org/wiki/Margaret_Hamilton_(software_engineer); https://en.wikipedia.org/wiki/Katherine_Johnson).

- When setting tile if the surface is cleaned and joints smoothed over what is necessary execessive grout can be removed from the joints, resulting in a weak installation. However, if epoxy grout is used and the surface and joints are not cleaned adequately when dry the epoxy can remain where it should not be visible resulting in unsightly finished job where cleaning epoxy grout off of surfaces of tile or stone, for example, slate, can be a very time consuming if possible at all. In this context there is a case for overanalyzing cleaning surfaces and smoothing joints after setting tile, with the caveat of knowing the material being used to avoid not under cleaning after setting tile or stone.

- If an individual accepted the story of "Betsy Ross sewed the first American flag" the individual would not examine the primary source of that story and vet the account to reach the fact that the Grand Union Flag predates the flag purported to be sewn by Betsy Ross, and further that the Grand Union Flag is identical to the East India Company Flag. (https://en.wikipedia.org/wiki/Grand_Union_Flag, https://en.wikipedia.org/wiki/East_India_Company). Further cursory analyzation of the story of Betsy Ross and the first American flag would reveal that what would become the renown Yale University received a gift from Elihu Yale (a former president of the East India Company who was sacked for aggrandizing) for the purposes of procuring a structure for the school. The specific topic of why the fledgling U.S. adopted a flag identical to the pre-existing East India Company flag cannot presently be overanalyzed.  It is simpler to find the official documentation of the meaning of each part of the Confederate States of America national flag than to find any official documentation as to the reason why the U.S. Grand Union Flag (the stripes on the current U.S. flag remain from the Grand Union Flag) is identical to the East India Company flag.

- The reputible institution Smithsonian recently published an article about the theory of human skin adaption (http://humanorigins.si.edu/evidence/genetics/human-skin-color-variation/modern-human-diversity-skin-color) where in pertinent parts states 
  >  The darker skin of peoples who lived closer to the equator was important in preventing folate deficiency. Measures of skin reflectance, a way to quantify skin color by measuring the amount of light it reflects, in people around the world support this idea. While UV rays can cause skin cancer, because skin cancer usually affects people after they have had children, it likely had little effect on the evolution of skin color because evolution favors changes that improve reproductive success.
  > 
  > There is also a third factor which affects skin color: coastal peoples who eat diets rich in seafood enjoy this alternate source of vitamin D. That means that some Arctic peoples, such as native peoples of Alaska and Canada, can afford to remain dark-skinned even in low UV areas. In the summer they get high levels of UV rays reflected from the surface of snow and ice, and their dark skin protects them from this reflected light.

which taken at face value could be considered to be a plausible theory. However, is the research scientist is not satisfied with the theory presented, they could ask the question if the theory is true and correct is the purported adaption occurring right now? That is, are the control groups of "dark-skinned" people who have resided in Scandanavian or Artic geographical regions for centuries, millenia, over generations, observably "lighter-skinned" and conversely, are "light-skinned" people who have resided in South Africa or Belize for centuries displaying the observable trait of "daker-skin", without the two purported two control groups mating, that is conclusively measurable and reproducible by means of the scientific method?

- Within the field of law, the discipline and codified principles of statutory construction, or determining what each word in a statute or adminsitrative regulation meant when enacted and the contemporaneous meaning, if any, is a very useful means of analyzing any subject matter or topic. If an individual were to accept the construct of "race" has been overanalyzed and that the topic has been settled they would by necessity need to ignore what modern science has to say ([There’s No Scientific Basis for Race—It's a Made-Up Label](https://www.nationalgeographic.com/magazine/2018/04/race-genetics-science-africa/)). Further analyzation of the topic of "race" would lead to the fact that if an individual were to _not_ to overanalyze the official documentation they might adopt the definition by the U.S. Census Bureau of "Black or African American", specifically the language "A person having origins in any of the black racial groups of Africa." (https://www.gpo.gov/fdsys/pkg/FR-1997-10-30/html/97-28653.htm) they could easily by negligence not ask the next logical question: how many "black racial groups of Africa" supposedly exist? Where, only after overanalyzation the answer is conclusively zero; as the only location where that language can be found, in any theory or discipline, is in that single document. Similarly for the definition of "White" in that document, without overanalyzation one might conclude that a geographical region named "North Africa" and "Middle East" actually exists, without asking precisely where "North Africa" and "Middle East" are, as neither is a de jure nation-state capable of being unambiguosly mapped using latitude and longitude; they exist merely as fictions in peoples minds who do not overanalyze topics. For example, no "Black" or "White" people officially exist in Germany or France. That particular thread is particularly in need of overanalyzation, where "White" and "Black" are classes invented relatively recently for a specific purpose: Project MUSE -  Working Definitions: Race, Ethnic Studies, and Early American Literature  "The idea of race came into being as a means of organizing social relations in order to establish and maintain political and economic domination." https://t.co/Elxc5Q64NL; the details of which (the invention of "race") have thoroughly vetted and is far too elaborate to detail here, though suffice to say, without overanalyzing the subbject matter would not have reached where there is no place to go other than concluding the fact that those inventions are one of the most, if not the most, horrendous contraption ever to be devised (which exists in individuals' minds when they do not overanalyze what they have been trained to not analyze at all, or to stop when others say stop because your analysis is contrary to what they believe, is unsettling to them, or they simply despise what you say "Goebbels was in favor of free speech for views he liked. So was Stalin. If you’re really in favor of free speech, then you’re in favor of freedom of speech for precisely the views you despise. Otherwise, you’re not in favor of free speech. -Noam Chomsky").

- Relevant to coding and overanalyzation: How do you know when you have reached `EOF`?  Without overanalyzation of what actually is occurring, as oppossed to what is specified (or not) to occur would not have found these two bugs https://bugzilla.mozilla.org/show_bug.cgi?id=1311823,  Without overanalyzation of what actually is occurring, as oppossed to what is specified (or not) to occur would not have found these two bugs https://bugzilla.mozilla.org/show_bug.cgi?id=1311823, https://bugs.chromium.org/p/chromium/issues/detail?id=816095.
https://bugs.chromium.org/p/chromium/issues/detail?id=816095.
Re the `MediaRecorder` specification (and actual implementation at browsers) and containers, codecs available with the code shipped with the browser, the fact is that Firefox supports only WebM and VP8 and Opus and Ogg. The official response from Firefox following request for including additional codecs is linked above. If that has changed, or is in the process of changing, kindly link to the document which states that.

guest271314:
@Pehrsons The Status `RESOLVED WONTFIX` at Webm support for h.264 codec https://bugzilla.mozilla.org/show_bug.cgi?id=1562862 is a clear indication that Mozilla appears to be reluctant to add support for container other than WebM and codecs other than VP8 and Opus (save for Ogg). If the codecs for `MediaRecorder` are or were intended to be fixed, static, then VP9 should not be supported either. *oogle Chrome and Chromium currently DO support `video/webm;codecs=h264`,  `video/webm;codecs=avc1`,  `video/x-matroska;codecs=h264` and  `video/x-matroska;codecs=avc1`, in spite of any purported limitation of the codecs that WebM was intended to support.

guest271314:
@Pehrsons For clarity https://groups.google.com/a/webmproject.org/forum/#!topic/webm-discuss/nJ39yu-CuKI ; https://bugzilla.mozilla.org/show_bug.cgi?id=1563140

guest271314:
> It seems reasonable to expect the video to be recorded with dynamically changing resolution, so that input frames are correctly represented in the output. The main issue I see though, is if recording into a container that doesn't support changing resolution dynamically. Should that be an error? Or should it be left to the UA to take an alternative approach, like scale, crop, etc?
> 
> I don't know the container landscape well enough to say whether, or which, containers support this, but I imagine there are some that don't. Others might know better.

@Pehrsons Interestingly see [Unexpected Error when appending webm files: mkvmerge throws error when width and height are different](https://gitlab.com/mbunkus/mkvtoolnix/issues/2582) at https://gitlab.com/mbunkus/mkvtoolnix/issues/2582#note_189360046

> This is not a bug. Matroska doesn't support changing pixel widths/heights in the middle of the stream.

Though that is evidently not the case as demonstrated by the output of `MediaRecorder` at Firefox. If that were/is the case per some inconspicuous note or detail at a controlling Matroska or WebM specification then Firefox is beyond the cutting edge as to writing Matroska-based WebM files and further the Firefox implementation needs to be the example of how such functionality IS achieved, which should lead to the relevant specifications being updated to reflect what is actually possible using the Matroska-based webm container.

jnoring:
> @Pehrsons Interestingly see [Unexpected Error when appending webm files: mkvmerge throws error when width and height are different](https://gitlab.com/mbunkus/mkvtoolnix/issues/2582) at https://gitlab.com/mbunkus/mkvtoolnix/issues/2582#note_189360046
> 
> > This is not a bug. Matroska doesn't support changing pixel widths/heights in the middle of the stream.

I haven't found any evidence this is true.  There's certainly nothing in the standard (https://www.matroska.org/technical/specs/index.html) that explicitly precludes this that I could find.  I'd also point out if you read Matroska's goals as a project (https://www.matroska.org/technical/whatis/index.html), I would fully expect them to support this.  Format changes mid-stream are possible (and even common) in many video codecs, including major ones like H.264 and VP8/VP9.

It _is_ true that Matroska has a container level width/height that gets set, and so it's possible that value could be misleading if a client is expecting no change to the format of the stream--but I also don't think it's unreasonable to set those values to whatever the video stream starts at, and expect clients to be able to understand stream-level format changes that roll through as they process data from the container.  

Also ffmpeg handles dynamic resolution changes just fine in mkv (I've been processing mkv's with dynamic format changes via ffmpeg for years now).

jnoring:
And one more comment: I'd argue mkvmerge does not support dynamic format changes, because they'd need to construct stream-level data to indicate a format change (completely possible without re-encoding, btw).  But that's a very different thing from mkv not handling dynamic format changes mid-stream.

guest271314:
@jnoring Re-opened the issue as could also not locate any specific language in the relevant specifications where Matroska container and derived WebM format restricts variable input stream `width` and `height`. Re-opened the issue; the author of `mkvmerge` acknowledges the above facts https://gitlab.com/mbunkus/mkvtoolnix/issues/2582#note_189460892

> You're right that the specs don't forbid it. However, this only works for video formats that have their width & height encoded at the bitstream level for each key frame as there's only one place those values can be stored at the container level. This isn't the case for all video codecs.
> I can remove this check for codecs that do contain that information, which include VP8, VP9, AV1, h.264 & h.265 (and with that the most popular codecs today).

guest271314:
@jnoring `mkvmerge` now supports variable width and height in same container (mkvmerge: allow appending video tracks with differing pixel dimensions https://gitlab.com/mbunkus/mkvtoolnix/commit/2548a7f3497940135e55be604a87b9495fdff94d). 

However, there are at least two Chromium bugs that prevent that browser from rendering variable width and height of input video track. One bug is linked at https://github.com/w3c/mediacapture-record/pull/172#issuecomment-506397242. Even when not using `MediaRecorder`, Chromium browser implementation of `HTMLMediaElement` does not dispatch `resize` event when video track pixel dimensions change within a `webm` file nor display the correct `videoWidth` and `videoHeight` of input file at `<video>` element https://bugs.chromium.org/p/chromium/issues/detail?id=983777. Until those bugs are fixed, even if this PR was included in the W3C `MediaRecorder` specification the current `HTMLVideoElement` implementation at Chromium will not display the resulting `webm` file variable width and height correctly at `<video>` within HTML (Issue 983777: HTMLVideoElement does not resize or display correct videoWidth or videoHeight https://bugs.chromium.org/p/chromium/issues/detail?id=983777).

alvestrand:
Closing pull request due to IPR issue (guest e pi is not allowed to make substantive contributions). Please continue discussion in bug #173.

Issue #179, opened 2019-08-16
Add method that allows user specification of which tracks go with what number in the container-dependent ordering


guest271314:
> I think the Right Solution is to make MediaRecorder have a method that allows you to specify which tracks go with what number in the container-dependent ordering.

This change means a method such as 

```
MediaRecorder.setTrackOrder({track: MediaRecorder.stream.getAudioTracks()[0], orderId:0});
MediaRecorder.setTrackOrder({track: MediaRecorder.stream.getVideoTracks()[0], orderId:1});
```

To the degree deemed to be controlling or relevant at this specification, track order language is found at https://w3c.github.io/mediacapture-main/

> The relative order of the tracks in the set is User Agent defined and the API will never put any requirements on the order.

However, after filing this issue at that specification, was referred to the mediacapture-record specification as owner of the context https://github.com/w3c/mediacapture-main/issues/611.

If a `MediaStream` contains both audio and video then the order described at the first sentence should be applied.

The reason for the change is for post-record editing, which becomes tedious when the order of the tracks is arbitrary.

Consider two `webm` files output by `MediaRecorder` at Chromium

```
$ ./mkvmerge -J 1.webm
{
  "attachments": [],
  "chapters": [],
  "container": {
    "properties": {
      "container_type": 17,
      "is_providing_timestamps": true,
      "muxing_application": "Chrome",
      "writing_application": "Chrome"
    },
    "recognized": true,
    "supported": true,
    "type": "Matroska"
  },
  "errors": [],
  "file_name": "1.webm",
  "global_tags": [],
  "identification_format_version": 12,
  "track_tags": [],
  "tracks": [
    {
      "codec": "Opus",
      "id": 0,
      "properties": {
        "audio_bits_per_sample": 32,
        "audio_channels": 1,
        "audio_sampling_frequency": 48000,
        "codec_id": "A_OPUS",
        "codec_private_data": "4f707573486561640101000080bb0000000000",
        "codec_private_length": 19,
        "default_track": true,
        "enabled_track": true,
        "forced_track": false,
        "language": "eng",
        "minimum_timestamp": 119000000,
        "number": 1,
        "uid": 40889044785409303
      },
      "type": "audio"
    },
    {
      "codec": "VP8",
      "id": 1,
      "properties": {
        "codec_id": "V_VP8",
        "codec_private_length": 0,
        "default_track": true,
        "display_dimensions": "1046x588",
        "display_unit": 0,
        "enabled_track": true,
[...]

PR #186, opened 2019-09-12
Add replaceStream to MediaRecorder

### LINKED ISSUE #167: Add replaceTrack method to MediaStream
Related:

- https://github.com/w3c/mediacapture-record/issues/4

- https://github.com/w3c/mediacapture-record/issues/166

- https://github.com/w3c/mediacapture-main/issues/575

- https://github.com/w3c/webrtc-pc/issues/2171

Given a `MediaStream` instance either created using `new MediaStream()` and/or by `getUserMedia()`, `HTMLMediaElement.captureStream()`, add a `replaceTrack()` method directly to the existing `MediaStream` which behaves similarly to `RTCRtpSender.replaceTrack()` https://w3c.github.io/webrtc-pc/#dom-rtcrtpsender-replacetrack

> NOTE
> 
> There is not an exact 1:1 correspondence between tracks sent by one RTCPeerConnection and received by the other. For one, IDs of tracks sent have no mapping to the IDs of tracks received. Also, replaceTrack changes the track sent by an RTCRtpSender without creating a new track on the receiver side; the corresponding RTCRtpReceiver will only have a single track, potentially representing **multiple sources of media stitched together**. Both addTransceiver and replaceTrack can be used to cause the same track to be sent multiple times, which will be observed on the receiver side as multiple receivers each with its own separate track. Thus it's more accurate to think of a 1:1 relationship between an RTCRtpSender on one side and an RTCRtpReceiver's track on the other side, matching senders and receivers using the RTCRtpTransceiver's mid if necessary. (emphasis added)

The primary use case for adding a `replaceTrack` method to `MediaStream` is for the ability to record media (for example following a call to `HTMLMediaElement.captureStream()`) with `MediaRecorder` where the `src` attribute of the `HTMLMediaElement` is changed (currently `MediaRecorder` stops recording when a track is added to the `MediaStream` being recorded by means of `src` of `<video>` or `<audio>` element changing). 

This https://github.com/guest271314/MediaFragmentRecorder/blob/webrtc-replacetrack/MediaFragmentRecorder.html is currently possible using `RTCPeerConnection` (proof of concept) though should be possible without the end-user having to manually create two `RTCPeerConnection` instances to achieve the requirement. Instead, the end-user should simply be able to execute `<MediaStreamInstance>.replaceTrack(withTrack)` where internally `replaceTrack(withTrack)` achieves the same functionality of `RTCRtpSender.replaceTrack()` 

> multiple sources of media stitched together

> which will be observed on the receiver side as multiple receivers each with its own separate track

Filing this issue before filing a PR for changes to the respective specification(s).

henbos:
Fixes #167

***
<a href="https://pr-preview.s3.amazonaws.com/henbos/mediacapture-record/pull/186.html" title="Last updated on Feb 15, 2021, 7:32 AM UTC (2e144c9)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/mediacapture-record/186/e17767b...henbos:2e144c9.html" title="Last updated on Feb 15, 2021, 7:32 AM UTC (2e144c9)">Diff</a>

### DIFF
--- a/MediaRecorder.bs
+++ b/MediaRecorder.bs
@@ -71,6 +71,7 @@ interface MediaRecorder : EventTarget {
   readonly attribute unsigned long audioBitsPerSecond;
 
   void start(optional unsigned long timeslice);
+  Promise&lt;void&gt; replaceStream(MediaStream newStream);
   void stop();
   void pause();
   void resume();
@@ -408,6 +409,46 @@ interface MediaRecorder : EventTarget {
       </tr>
     </tbody>
   </table>
+  </dd>
+
+  <dt><dfn method for="MediaRecorder"><code>replaceStream(MediaStream newStream)</code></dfn></dt>
+  <dd>
+  When a {{MediaRecorder}} object&#8217;s {{replaceStream()}} method is invoked,
+  the UA MUST run the following steps:
+  <ol>
+    <li>Let |recorder| be the {{MediaRecorder}} object on which the method was
+    invoked.</li>
+    <li>Let |newStream| be the method's first argument.</li>
+    <li>Let |newTracks| be the set of {{live}} tracks in |newStream|'s
+    [=track set=].</li>
+    <li>Let |stream| be the value of |recorder|'s {{stream}} attribute.</li>
+    <li>Let |tracks| be the set of {{live}} tracks in |stream|'s
+    [=track set=].</li>
+    <li>If the number of audio tracks in |newTracks| is different than the
+    number of audio tracks in |tracks|, or the number of video tracks in
+    |newTracks| is different than the number of video tracks in |tracks|, return
+    a promise {{rejected}} with a newly created {{InvalidModificationError}} and
+    abort these steps.</li>
+    <li>Let |p| be a new promise and run the following steps in parallel:</li>
+    <ol>
+      <li>If data is currently being gathered, replace the source stream being
+      gathered to be |newStream|, recording from |newTracks| instead of the
+      previous stream and tracks. See {{play()}} for details on what to do with
+      the stream and tracks being recorded (such as what to do if the recorded
+      stream's [=track set=] is modified) and stop listening for these things on
+      the stream no longer being recorded.</li>
+      <div class="note">This replaces the source of frames being gathered and
+      should be a seamless operation.</div>
+      <li>Queue a task, using the DOM manipulation task source, that runs the
+      following steps:</li>
+      <ol>
+        <li>Set |recorder|'s {{MediaRecorder/stream}} attribute to
+        |newStream|.</li>
+        <li>Resolve |p| with <code>undefined</code>.</li>
+      </ol>
+    </ol>
+    <li>return |p|.</li>
+  </ol>
   </dd>
 
   <dt><dfn method for="MediaRecorder"><code>stop()</code></dfn></dt>


jan-ivar:
cc @Pehrsons

MediaRecorder.bs
```diff
@@ -410,6 +411,46 @@ interface MediaRecorder : EventTarget {
...
+    <li>Let |tracks| be the set of {{live}} tracks in |stream|'s
+    [=track set=].</li>
+    <li>If the number of audio tracks in |newTracks| is different than the
+    number of audio tracks in |tracks|, or the number of video tracks in
+    |newTracks| is different than the number of video tracks in |tracks|, return
+    a promise {{rejected}} with a newly created {{InvalidModificationError}} and
+    abort these steps.</li>
+    <li>Let |p| be a new promise and run the following steps in parallel:</li>
+    <ol>
+      <li>If data is currently being gathered, replace the source stream being
```
jan-ivar, line 434: What happens if data is not currently being recorded?

guest271314: @jan-ivar `MediaRecorder` state being active, inactive or paused should not affect replacing the principal `MediaStream`.

MediaRecorder.bs
```diff
@@ -410,6 +411,46 @@ interface MediaRecorder : EventTarget {
...
+  <ol>
+    <li>Let |recorder| be the {{MediaRecorder}} object on which the method was
+    invoked.</li>
+    <li>Let |newStream| be the method's first argument.</li>
+    <li>Let |newTracks| be the set of {{live}} tracks in |newStream|'s
+    [=track set=].</li>
+    <li>Let |stream| be the value of |recorder|'s {{stream}} attribute.</li>
+    <li>Let |tracks| be the set of {{live}} tracks in |stream|'s
+    [=track set=].</li>
+    <li>If the number of audio tracks in |newTracks| is different than the
```
guest271314, line 427: @henbos There should be a means other than an error to notify the user that an audio or video track needs to be added to the new `MediaStream` both before `start()` is executed and ideally also when `start()` is executed, to preserve that instance of `MediaRecorder`. `streamsMatch()` returning a Boolean or track order and track kind in the old and new tracks. `matchStreams()` to add the necessary track kind (transparent, #000000 frames and silence) to new stream when not present. Both `streamsMatch` and `matchStream` can be arguments passed to `replaceStream()` as an object instead of methods, and be return values from `replaceStream()`.

alvestrand: Under the proposed API, having a replaceStream() being rejected doesn't affect the recorder.

henbos: @alvestrand That's correct. Upon rejection the application can decide whether to stop the recording or to continue recording with the old track. Would you prefer that failing also triggered stop()?

@guest271314 I don't think the API should silently try to patch the stream so that it will get accepted; what type of tracks you would add e.g. when video is missing is application-specific. Better to inform the application that they're doing something unexpected. Discussing additional APIs to help the application achieve matching streams on the other hand could be discussed separately if you file another issue.

MediaRecorder.bs
henbos: @jan-ivar If media is not being recorded this step is skipped, and we immediately jump back and replace the stream attribute and resolve the promise.

jan-ivar:
Discussed this with @Pehrsons and we have some concerns with this API.

While browsers today appear limited to recording only one video track and one audio track, there's no such limitation in the spec, so we think we need to consider how to handle several tracks.

Say someone records a stream like this:
```js
const stream = new MediaStream([videoTrack, englishAudioTrack, frenchAudioTrack]);
const recorder = new MediaRecorder(stream);
recorder.start();
```
How would they replace just the french audio? Tracks in a MediaStream unfortunately are unordered.

So we'd like to propose this instead:
```js
await recorder.replaceTrack(frenchAudioTrack, directorsCommentaryTrack);
```
In addition, we propose to no longer stop the recorder if a stream's track-set changes, because it no longer makes sense to react to those. Instead, we'd copy the track-set in `start()`, and only stop on the first track `ended`.

If the tracks' isolation properties differ, `replaceTrack` would reject with `SecurityError`.

If`stop()` followed by `start()` is called again then the stream can be examined anew.

guest271314:
@jan-ivar 

> How would they replace just the french audio?

That is in fact a consideration that suggested be included in this PR at https://github.com/w3c/mediacapture-record/issues/167#issuecomment-531060600. In brief, by providing a means to set the order the tracks. Further, by providing a means to parse media files. 

> So we'd like to propose this instead:
> 
> `await recorder.replaceTrack(frenchAudioTrack, directorsCommentaryTrack);`

"instead" is not necessary language or intent. There is no reason that both `replaceStream()` and `replaceTrack()` cannot be specified and implemented.

> In addition, we propose to no longer stop the recorder if a stream's track-set changes, because it no longer makes sense to react to those. 

That essentially would resolve the issue and make the use case viable. 

> Instead, we'd copy the track-set in start(), and only stop on the first track ended. 

That would be an issue for interoperability re Chromium, Chrome https://github.com/w3c/mediacapture-fromelement/issues/78.

> If the tracks' isolation properties differ, replaceTrack would reject with SecurityError.

AFAIK `isolationchange` is not implement, correct? Do you mean `unmute` event?

guest271314:
@jan-ivar Re "instead" see https://github.com/w3c/mediacapture-record/issues/167#issuecomment-526226683 where @henbos mentioned specifying both `replaceStream()` and `replaceTrack()`. 

Ideally the Media Capture and Streams (main) will eventually abandon the concept of tracks not being ordered https://github.com/w3c/mediacapture-main/issues/611,  https://github.com/w3c/mediacapture-record/issues/179 due to the fact of the issues arising when post-production on the WebM file(s) are performed. 

For this input (some files may have audio track without video track or video track without audio track) 

> https://upload.wikimedia.org/wikipedia/commons/6/6e/Micronesia_National_Anthem.ogg#t=0,2
> https://upload.wikimedia.org/wikipedia/commons/a/a4/Xacti-AC8EX-Sample_video-001.ogv#t=0,4
> https://mirrors.creativecommons.org/movingimages/webm/ScienceCommonsJesseDylan_240p.webm#t=10,20
> https://mirrors.creativecommons.org/movingimages/webm/ASharedCulture_480p.webm#t=22,26
> https://nickdesaulniers.github.io/netfix/demo/frag_bunny.mp4#t=55,60
> https://raw.githubusercontent.com/w3c/web-platform-tests/master/media-source/mp4/test.mp4#t=0,5
> https://commondatastorage.googleapis.com/gtv-videos-bucket/sample/ForBiggerBlazes.mp4#t=0,5
> https://commondatastorage.googleapis.com/gtv-videos-bucket/sample/ForBiggerJoyrides.mp4#t=0,5
> https://commondatastorage.googleapis.com/gtv-videos-bucket/sample/ForBiggerMeltdowns.mp4#t=0,6

the required mapping to merge the tracks, after creating the corresponding potentially omitted video or audio track 

```
[
  {
    "audio": 1,
    "video": 0
  },
  {
    "audio": 0,
    "video": 1
  },
  {
    "audio": 1,
    "video": 0
  },
  {
    "audio": 0,
    "video": 1
  },
  {
    "audio": 0,
    "video": 1
  },
  {
    "audio": 0,
    "video": 1
  },
  {
    "audio": 0,
    "video": 1
  },
  {
    "audio": 0,
    "video": 1
  },
  {
    "audio": 0,
    "video": 1
  }
]
```

though is not guranteed to be in that same order for each run of the code. To merge with `mkvmerge` `--append-to` mapping is

`1:0:0:1,1:1:0:0,2:1:1:0,2:0:1:1,3:0:2:1,3:1:2:0,4:0:3:0,4:1:3:1,5:0:4:0,5:1:4:1,6:0:5:0,6:1:5:1,7:0:6:0,7:1:6:1,8:0:7:0,8:1:7:1`.

Result https://plnkr.co/edit/8J61Rw?p=info.

Even when WebM file produced by Chromium and Mozilla browsers have the same track order they still cannot be merged; an error is thrown, due to the vastly different parameters passed to the encoder. 

Thus, there is still must ground to be covered for any sort of interoperability, if that term or concept is in fact being considered at all during this specification write-up. Setting consistent track order should eventually be addressed. There is no compelling reason to not have consistent track order, especially within the purview of `MediaRecorder`.

jan-ivar:
https://github.com/w3c/mediacapture-main/issues/611 was closed. I have no hopes we'll define track order in MediaStream.

MediaRecorder could do what [web audio does](https://github.com/WebAudio/web-audio-api/pull/1811/) to make order deterministic, though uuid *generation* being random, it's not super practical.

I see no value in `replaceStream` if we have `replaceTrack`—the former can be implemented by the latter. In general we've been moving away from stream APIs, to avoid action at a distance.

> AFAIK isolationchange is not implement, correct? Do you mean unmute event?

No, even though no browser currently exposes that or the [isolated](https://w3c.github.io/webrtc-identity/identity.html#dom-mediastreamtrack-isolated) attribute, tracks still have "isolation properties" in all browsers, which [can be observed](https://jsfiddle.net/jib1/1kz9hfaL/) e.g. with cross-origin content. We have to guard cross-origin content.

guest271314:
@jan-ivar 

> w3c/mediacapture-main#611 was closed. I have no hopes we'll define track order in MediaStream.

Do not entertain hope at all here. As a practical matter must defer to your experience and judgment re `MediaStream` specification.

> I see no value in `replaceStream` if we have `replaceTrack`—the former can be implemented by the latter. In general we've been moving away from stream APIs, to avoid action at a distance.

Then define `replaceTrack` for `MediaRecorder` yourself. The original proposal is for `replaceTrack` as the mechanism that you apparently single-handedly pushed through WebRTC specification is well-suited to achieve the requirement of the use case, as demonstrated at at least four branches at code to work around current `MediaRecorder` implementation https://github.com/guest271314/MediaFragmentRecorder/branches/all?utf8=✓&query=webrtc-replacetrack, and several issues filed at various specification repositories at GitHub, et al.

At the front-end, since @henbos evidently has interest in extending `MediaRecorder` with `replaceStream` would rather have that specified than remaining with the same specification. Do gather your point. 

Even if `replaceStream` is specified that should not stop you and interested parties from specifying `replaceTrack`. The proof of concept will be vetted by the language of the specification and borne out by the implementations therefrom, to the degree that implementers actually follow specifications until they decide to do otherwise. AFAICT no individual or entity is or can stop you from filing a PR to add `replaceTrack` to `MediaRecorder` right now, correct?

Should @henbos be asked to substitute `replaceTrack` for `replaceStream` and abandon their own proposal for the reasons you raised, for the purpose of having a uniform, cohesive concept moving forward? 

How do you suggest to proceed?

guest271314:
@jan-ivar It must be noted here that all branches at https://github.com/guest271314/MediaFragmentRecorder besides master https://github.com/guest271314/MediaFragmentRecorder/blob/master/MediaFragmentRecorder.html were written due to the fact that Chromium crashed the browser when trying to record a variable resolution video where HTML `<video>` element `src` was set to `MediaSource`. Initially considered the bug to be `MediaSource`, though that is apparently not the case https://bugs.chromium.org/p/chromium/issues/detail?id=1001948, and Chromium **_still_** crashes the browser when such recording is attempted. Mozilla Firefox and Nightly had and have no issues. 

The meaning of the above statement of facts is that the goal was and is always to write code that ouputs the same result at Mozilla browsers and Chromium, Chrome browsers. Otherwise would have simply used Mozilla for the purpose of recording media fragments potentially having variable resolution. 

Therefore, whatever specification change (if that actually does occur) arises from https://github.com/w3c/mediacapture-record/issues/167 should ultimately be implemented at browsers in the same manner, else the words in the specification will be void as to the true intent of the numerous issues that have filed relevant to `MediaRecorder` having the ability to record multiple tracks from varying sources.

Perhaps it is not possible for Mozilla and Chromium, Chrome authors to actually work together to achieve a single aim.  We shall see.

guest271314:
@henbos What say you regarding the points made by @jan-ivar at https://github.com/w3c/mediacapture-record/pull/186#issuecomment-531524506 ?

> I see no value in `replaceStream` if we have `replaceTrack`—the former can be implemented by the latter. In general we've been moving away from stream APIs, to avoid action at a distance.

Pehrsons:
> Discussed this with @Pehrsons and we have some concerns with this API.
> 
> While browsers today appear limited to recording only one video track and one audio track, there's no such limitation in the spec, so we think we need to consider how to handle several tracks.
> 
> Say someone records a stream like this:
> 
> ```js
> const stream = new MediaStream([videoTrack, englishAudioTrack, frenchAudioTrack]);
> const recorder = new MediaRecorder(stream);
> recorder.start();
> ```
> 
> How would they replace just the french audio? Tracks in a MediaStream unfortunately are unordered.
> 
> So we'd like to propose this instead:
> 
> ```js
> await recorder.replaceTrack(frenchAudioTrack, directorsCommentaryTrack);
> ```

I'll just comment on this that one can trivially polyfill the functionality of `replaceStream` with `replaceTrack` (modulo replacing the recorder's `stream` attribute). But polyfilling `replaceTrack` with `replaceStream` becomes hacky as one would have to replace one track at a time.

> In addition, we propose to no longer stop the recorder if a stream's track-set changes, because it no longer makes sense to react to those. Instead, we'd copy the track-set in `start()`, and only stop on the first track `ended`.
> 
> If the tracks' isolation properties differ, `replaceTrack` would reject with `SecurityError`.
> 
> If`stop()` followed by `start()` is called again then the stream can be examined anew.

I thought some more about the need to for the `replace` method to return a promise. The only argument I've heard in favor of returning a promise is that then you know you can dispose of the old tracks without risking ending them in the recording before the new ones take effect.

I think this is unnecessary. The MediaRecorder could guarantee that `recorder.replaceTrack(oldTrack, newTrack); oldTrack.stop();` replaces the track successfully as long as the track is not ended.

To stretch things a bit, the recorder could guarantee that the track is replaced successfully as long as the recorder is not in the `inactive` state.

This would help applications handle concatenating/swapping tracks when the application itself is not in charge of the lifetime of those tracks, like HTMLMediaElement's `captureStream()`.

In fact, if one of two tracks the recorder is recording has ended, there's nothing to say in the spec that the recording of the ended track (let's call it the output track) has also ended. The spec says
> If **_all_** recorded tracks become ended, then stop gathering data, and queue a task (...) Inactivate the recorder

(emphasis mine)

Though this highlights a race that would need to be fixed, in the case of:
```
const rec = new MediaRecorder(stream);
rec.start();
await new Promise(r => rec.onstart);
stream.getTracks().forEach(t => t.stop()); // stops gathering data and queues a task to inactivate
rec.replaceTrack(stream.getTracks()[0], newTrack); // rec.state is not inactive yet -- should this throw? -- or abort the task queued above?
```
Perhaps that'd be the reason we need to return a promise.

jan-ivar:
@pehrsons If we can make those guarantees then I agree we shouldn't return a promise. Only if the asynchronous part can fail would a promise be needed. I don't think that's the case here. 

`replaceTrack` should probably throw if either track is `ended`. That would avoid the race you mention.

jan-ivar:
Also, with `replaceTrack` there's inherently no way to add additional tracks to a recording, whereas with `replaceStream` you'd be forgiven for not knowing that this limitation exists (it's in the prose).

Pehrsons:
> @Pehrsons If we can make those guarantees then I agree we shouldn't return a promise. Only if the asynchronous part can fail would a promise be needed. I don't think that's the case here.
> 
> `replaceTrack` should probably throw if either track is `ended`. That would avoid the race you mention.

@jan-ivar Personally I like that it solved the track-lifetime problem more than I'd be annoyed by having to solve the race.

Consider recording a stream from HTMLMediaElement.captureStream(). You could pipe an audio track through webaudio to be in full control of its lifetime. Then even when a video track from the media element ended before you managed to replace it with another (which if you managed to do it in time, would chop off a bit of the track's end), you can still do that after-the-fact. Or with a gUM track in case a user killed it through the browser's chrome.

MediaRecorder.bs
guest271314: What should occur when the use case is one audio track with multiple video tracks being replaced? Should a new `MediaStream` object be created for each video track replacement with the single audio track being added to the replacement `MediaStream` from the replaced `MediaStream`? 

What are the reasons for proposing `replaceStream()` instead of `replaceTrack()`?

guest271314:
@jan-ivar 

> Also, with `replaceTrack` there's inherently no way to add additional tracks to a recording, whereas with `replaceStream` you'd be forgiven for not knowing that this limitation exists (it's in the prose).

If gather the statement context correctly, yes there is. Given the case of recording multiple tracks at a single `<video>` element from `HTMLVideElement.captureStream()` execute `captureStream()` at a before any playback begins, when `src` of `<video>` is changed the `MediaStreamTrack`s from the new media source are added to the original `MediaStream` from `captureStream()` and `addtrack` event is dispatched, where within the event handler `replaceTrack()` can be executed for each audio or video track, a 1:1 swap for new track:original track N times, where the `MediaStream` contains at most two tracks throughout the process https://github.com/guest271314/MediaFragmentRecorder/blob/webrtc-replacetrack-htmlmediaelement-capturestream-addtrack/MediaFragmentRecorder.html.

guest271314:
@Pehrsons 

> Consider recording a stream from HTMLMediaElement.captureStream(). You could pipe an audio track through webaudio to be in full control of its lifetime. Then even when a video track from the media element ended before you managed to replace it with another (which if you managed to do it in time, would chop off a bit of the track's end), you can still do that after-the-fact. Or with a gUM track in case a user killed it through the browser's chrome.

That approach is possible with both audio and video. The concept of this code is an _infinite_ media stream initialized to `#000000` video frames and audio silence (utilizing `AudioWorklet`, `Worker`, and `canvas.captureStream()`) where `MediaStreamTrack`s can be added to the `MediaStream` at any time. When there is a "gap" between track additions or swaps the initial audio or video track is used for that respective track data. The entire procedure should be capable of being recorded by `MediaRecorder` with the separate capability to seek backwards or forwards for infinity given ["an idealized computer with unlimited resources"](https://en.wikipedia.org/wiki/Kurt_Gödel#Incompleteness_theorem) (for brevity citing Wikipedia), though during testing the code the tab inevitably crashes (there should be some means to reduce or memory usage for the APIs used) https://plnkr.co/edit/Axkb8s?p=info.

The original issue proposing using the same or similar language to WebRTC `RTCRtpSender.replaceTrack()`. Ideally, neither a track not currently having data, nor the source `<video>` pausing, nor the track or `MediaStreamTrack` ending should be the cause of `MediaRecorder` stopping. `#000000` video frames and audio silence can be recorded by default until the user explicitly executes `stop()`.

Pehrsons:
> @jan-ivar
> 
> > Also, with `replaceTrack` there's inherently no way to add additional tracks to a recording, whereas with `replaceStream` you'd be forgiven for not knowing that this limitation exists (it's in the prose).
> 
> If gather the statement context correctly, yes there is.

I think with "Add additional tracks" he means going from recording two tracks to three, or from 20 to 25, or the like. That would seem possible from just looking at the `replaceStream` API, even though it's not.

> The original issue proposing using the same or similar language to WebRTC `RTCRtpSender.replaceTrack()`. Ideally, neither a track not currently having data, nor the source `<video>` pausing, nor the track or `MediaStreamTrack` ending should be the cause of `MediaRecorder` stopping. `#000000` video frames and audio silence can be recorded by default until the user explicitly executes `stop()`.

The spec says to end the recording when all tracks have ended. Changing that would break existing applications that depend on this feature so it's not really an option.

alvestrand:
Closing because of lack of developer interest.

realies:
I would prefer not to use a `RTCPeerConnection` workaround. @alvestrand, can we reopen the issue, please?

ozturkkl:
Both RTCPeerConnection and using with MediaStreamTrackProcessor add overhead and reduce the FPS of the resulting recording. Is there any other solution that I'm not aware of? This is proving to be a massive pain point. I just want to be able to record a camera and if the feed drops, make sure it gets picked up again by re-grabbing the user media and adding the track to the existing recorder.

realies:
Not only the FPS but the bitrate becomes terrible and uncontrollable with the suggested workaround. Swapping video tracks should be possible like swapping audio tracks is.

The specification should not be driven by 'developer interest' but by what makes sense.

PR #187, opened 2019-09-17
Add replaceTrack method to MediaRecorder.

jan-ivar:
Alternative fix for https://github.com/w3c/mediacapture-record/issues/167.

***
<a href="https://pr-preview.s3.amazonaws.com/jan-ivar/mediacapture-record/pull/187.html" title="Last updated on Feb 15, 2021, 7:32 AM UTC (82be3e1)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/mediacapture-record/187/e17767b...jan-ivar:82be3e1.html" title="Last updated on Feb 15, 2021, 7:32 AM UTC (82be3e1)">Diff</a>

### DIFF
--- a/MediaRecorder.bs
+++ b/MediaRecorder.bs
@@ -71,7 +71,7 @@ interface MediaRecorder : EventTarget {
   readonly attribute unsigned long audioBitsPerSecond;
 
   void start(optional unsigned long timeslice);
-  void replaceTrack(MediaStreamTrack track, MediaStreamTrack withTrack);
+  Promise&lt;void&gt; replaceTrack(MediaStreamTrack track, MediaStreamTrack withTrack);
   void stop();
   void pause();
   void resume();
@@ -129,6 +129,10 @@ interface MediaRecorder : EventTarget {
     <li>Let |recorder| be a newly constructed {{MediaRecorder}} object.</li>
     <li>Let |recorder| have a <dfn>\[[Tracks]]</dfn> internal slot, initialized
     to an empty list.</li>
+    <li>Let |recorder| have a <dfn>\[[ReplaceAtOnceList]]</dfn> internal slot,
+    initialized to an empty list.</li>
+    <li>Let |recorder| have a <dfn>\[[ReplacePromise]]</dfn> internal slot,
+    initialized to <code>null</code>.</li>
     <li>Let |recorder| have a <dfn>\[[ConstrainedMimeType]]</dfn> internal slot,
     initialized to the empty string.</li>
     <li>Let |recorder| have a <dfn>\[[ConstrainedBitsPerSecond]]</dfn> internal
@@ -410,30 +414,55 @@ interface MediaRecorder : EventTarget {
     invoked.</li>
     <li>Let |track| be the method's first argument.</li>
     <li>Let |withTrack| be the method's second argument.</li>
-    <li>If |recorder|'s {{state}} is {{inactive}}, throw an
-    {{InvalidStateError}} {{DOMException}} and abort these steps.</li>
-    <li>If |track| is not found in |recorder|'s [=[[Tracks]]=], throw a
-    {{NotFoundError}} {{DOMException}} and abort these steps.</li>
-    <li>If |track| is {{ended}}, or |withTrack|'s {{kind}} differs from
-    |track|'s {{kind}}, then throw a {{InvalidModificationError}}
-    {{DOMException}} and abort these steps.</li>
-    <li>If the <a>isolation properties</a> of |track| or |withTrack| disallow
-    access from |recorder|, throw a {{SecurityError}} {{DOMException}} and abort
+    <li>If |recorder|'s {{state}} is {{inactive}}, return a promise rejected
+    with a newly created {{InvalidStateError}} {{DOMException}} and abort these
+    steps.</li>
+    <li>If |track| is not found in |recorder|'s [=[[Tracks]]=], return a promise
+    rejected with a newly created {{NotFoundError}} {{DOMException}} and abort
     these steps.</li>
+    <li>If |track| is {{ended}}, or |withTrack|'s {{kind}} differs from
+    |track|'s {{kind}}, then return a promise rejected with a newly created
+    {{InvalidModificationError}} {{DOMException}} and abort these steps.</li>
+    <li>If the <a>isolation properties</a> of |track| or |withTrack| disallow
+    access from |recorder|, return a promise rejected with a newly created
+    {{SecurityError}} {{DOMException}} and abort these steps.</li>
     <li>If the User Agent cannot record |withTrack| in place of |track| using
-    the current configuration, then throw a {{NotSupportedError}}
-    {{DOMException}} and abort these steps.</li>
+    the current configuration, then return a promise rejected with a newly
+    created {{NotSupportedError}} {{DOMException}} and abort these steps.</li>
     <li>Let |session| be the ongoing recording session corresponding to the
     first occurance of |track| in |recorder|'s [=[[Tracks]]=].</li>
     <li>Replace the first occurance of |track| in |recorder|'s [=[[Tracks]]=]
     with |withTrack|.</li>
-    <li>return <code>undefined</code>, and run the remaining step in parallel.
+    <li>Add |session| and |withTrack| as a pair to the list in |recorder|'s
+    [=[[ReplaceAtOnceList]]=] slot.</li>
+    <li>If |recorder|'s [=[[ReplacePromise]]=] slot is not <code>null</code>,
+    return it, and abort these steps.</li>
+    <li>Let |p| be a new promise.</li>
+    <li>Set |recorder|'s [=[[ReplacePromise]]=] slot to |p|, and queue
+    a task, using the DOM manipulation task source, that runs the following
+    steps:</li>
+    <ol>
+      <li>If |recorder|'s {{state}} attribute is {{inactive}}, resolve |p|, and
+      abort these steps.</li>
+      <li>Let |replaceList| be |recorder|'s [=[[ReplaceAtOnceList]]=], and set
+      |recorder|'s [=[[ReplaceAtOnceList]]=] to an empty list.</li>
+      <li>Set |recorder|'s [=[[ReplacePromise]]=] slot to <code>null</code>, and
+      run the remaining steps in parallel:</li>
+      <li>For each recording |session| and |withTrack| pair in |replaceList|,
+      instantly switch |session| to recording |withTrack| in place of what
+      |session| is currently recording.
+      <div class="note">This replaces the source of frames being gathered and
+      SHOULD be a seamless operation. This SHOULD happen at the same time for
+      all sessions in |recorder|'s [=[[ReplaceAtOnceList]]=].</div>
+      </li>
+      <li>Resolve |p|.
+      <div class="note">|p| MUST NOT be rejected for any reason, since track
+      replacement is observably synchronous in the API by calling |replaceTrack|
+      multiple times. The promise exists to simplify learning when replaced
+      tracks can be stopped without impacting the recording.</div></li>
+    </ol>
     </li>
-    <li>For |session|, instantly switch to recording |withTrack| in place of
-    |track|.
-    <div class="note">This replaces the source of frames being gathered and
-    SHOULD be a seamless operation.</div>
-    </li>
+    <li>return |p|.</li>
   </ol>
   </dd>
 
--- a/MediaRecorder.bs
+++ b/MediaRecorder.bs
@@ -58,7 +58,7 @@ at regular intervals.
 [Exposed=Window,
  Constructor(MediaStream stream, optional MediaRecorderOptions options)]
 interface MediaRecorder : EventTarget {
-  readonly attribute MediaStream stream;
+  attribute MediaStream stream;
   readonly attribute DOMString mimeType;
   readonly attribute RecordingState state;
   attribute EventHandler onstart;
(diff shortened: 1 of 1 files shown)


jan-ivar:
Based on https://github.com/w3c/mediacapture-record/pull/186#issuecomment-531421759 and as described in [slide](https://docs.google.com/presentation/d/1Xprz0ElNTUWRDdX6kyp11cZ4ItTvUFUmzoXsKZdXxuU/edit?ts=5d5d8e92#slide=id.g634b2fa483_43_0). Also fixed up isolation properties to be per track.

MediaRecorder.bs
```diff
@@ -410,6 +401,42 @@ interface MediaRecorder : EventTarget {
...
+  When a {{MediaRecorder}} object&#8217;s {{replaceTrack()}} method is invoked,
+  the UA MUST run the following steps:
+  <ol>
+    <li>Let |recorder| be the {{MediaRecorder}} object on which the method was
+    invoked.</li>
+    <li>Let |track| be the method's first argument.</li>
+    <li>Let |withTrack| be the method's second argument.</li>
+    <li>If |recorder|'s {{state}} is {{inactive}}, throw an
+    {{InvalidStateError}} {{DOMException}} and abort these steps.</li>
+    <li>If |track| is not found in |recorder|'s [=[[Tracks]]=], throw a
```
guest271314: Does that mean that the individual `MediaStreamTrack` must be referenced directly? E.g., either 

`let track = mediaStreamDestination.stream.getAudioTracks() // track is passed to recorder, still the same track`

or 

`let track = recorder.stream.getAudioTracks()[0] // the same track passed to recorder referenced now using recorder.stream` 

?

henbos:
Promise to add a promise

henbos:
Promise me no promises?

guest271314:
@jan-ivar @henbos Reviewing the TPAC minutes could not locate mention of `replaceStream()` or `replaceTrack()`. Were the PRs in this repository for the proposals brought up and discussed formally or informally? If so, what were the conclusion reached?

jan-ivar:
It was discussed [here](https://www.w3.org/2019/09/18-webrtc-minutes.html#item09). I felt there was support for progress, though no final consensus as I recall.

guest271314:
@jan-ivar  What does that ultimately mean as a practical matter?

MediaRecorder.bs
jan-ivar: Yes, to know which track to replace in the general case of more than one track of a kind.

In the simple case (of one video one audio) a polyfill would be:
```js
recorder.replaceStream = async function (stream) {
  const old = recorder.stream;
  await recorder.replaceTrack(old.getAudioTracks()[0], stream.getAudioTracks()[0]);
  await recorder.replaceTrack(old.getVideoTracks()[0], stream.getVideoTracks()[0]);
  recorder.stream = stream;
}
```

jan-ivar:
@Pehrsons PTAL. The concern raised at TPAC was replacing multiple tracks in concert, and whether returning a promise might simplify knowing when it is safe to stop all replaced tracks.

Here's my attempt. Note that MediaRecorder is a different beast from e.g. RTCPeerConnection when it comes to state. Changes are often synchronously observable, and asynchronous operations are not serialized (no operations chain).

MediaRecorder.bs
```diff
@@ -233,9 +241,9 @@ interface MediaRecorder : EventTarget {
     <li>If the value of |recorder|'s {{state}} attribute is not {{inactive}},
     throw an {{InvalidStateError}} {{DOMException}} and abort these steps.</li>
 
-    <li>If the <a>isolation properties</a> of |stream| disallow access from
-    |recorder|, throw a {{SecurityError}} {{DOMException}} and abort these
-    steps.</li>
+    <li>For each track, |track|, in |tracks|, if the <a>isolation properties</a>
```
Pehrsons, line 244: For some reason I can't leave review comments in Firefox, so now I'm doing it in Chrome.

`<a>isolation properties</a>` links to [isolated-media-streams](https://w3c.github.io/webrtc-identity/identity.html#isolated-media-streams) which is mostly talking about MediaStreams. Confusing at best.

MediaRecorder.bs
```diff
@@ -292,11 +302,11 @@ interface MediaRecorder : EventTarget {
...
       <code>undefined</code>.</li>
 
-      <li>If at any point |stream|'s <a>isolation properties</a> change so
-      that {{MediaRecorder}} is no longer allowed access to it, the UA MUST
-      immediately stop gathering data, discard any data that it has gathered,
-      and queue a task, using the DOM manipulation task source, that runs the
-      following steps:
+      <li>If at any point, the <a>isolation properties</a> of a track being
+      recorded change so that {{MediaRecorder}} is no longer allowed access a
+      track, the UA MUST immediately stop gathering all data, discard any data
```
Pehrsons, line 307: "if ... the isolation properties of *a track* ... change so that MediaRecorder is no longer allowed access *a track*"

1) Missing a "to" somewhere, either "allowed to access a track" or "allowed access to a track".
2) *a track* and then *a track* again -- make it clear that we're talking about the same track?

MediaRecorder.bs
```diff
@@ -365,15 +361,14 @@ interface MediaRecorder : EventTarget {
   <p>Note that {{stop()}}, {{requestData()}}, and {{pause()}} also affect the
   recording behavior.</p>
 
-  <p>The UA MUST record {{MediaRecorder/stream}} in such a way that the original
-  Tracks can be retrieved at playback time. When multiple {{Blob}}s are returned
+  <p>The UA MUST record track in such a way that they can be retrieved
```
Pehrsons, line 364: `s/track/|tracks|/` ?
"retrieved individually at playback time" seems vague, but is no worse than the status quo.

MediaRecorder.bs
```diff
@@ -410,6 +405,67 @@ interface MediaRecorder : EventTarget {
...
+    invoked.</li>
+    <li>Let |track| be the method's first argument.</li>
+    <li>Let |withTrack| be the method's second argument.</li>
+    <li>If |recorder|'s {{state}} is {{inactive}}, return a promise rejected
+    with a newly created {{InvalidStateError}} {{DOMException}} and abort these
+    steps.</li>
+    <li>If |track| is not found in |recorder|'s [=[[Tracks]]=], return a promise
+    rejected with a newly created {{NotFoundError}} {{DOMException}} and abort
+    these steps.</li>
+    <li>If |track| is {{ended}}, or |withTrack|'s {{kind}} differs from
```
Pehrsons, line 423: I think it makes sense to allow it to be ended as not all sources of tracks leave control of a track's lifetime to the application.

MediaRecorder would end the recording only when all tracks have ended, so if only some are ended it could encode silence for audio and simply nothing for video.

MediaRecorder.bs
```diff
@@ -410,6 +405,67 @@ interface MediaRecorder : EventTarget {
...
+      |session| is currently recording.
+      <div class="note">This replaces the source of frames being gathered and
+      SHOULD be a seamless operation. This SHOULD happen at the same time for
+      all sessions in |recorder|'s [=[[ReplaceAtOnceList]]=].</div>
+      </li>
+      <li>Resolve |p|.
+      <div class="note">|p| MUST NOT be rejected for any reason, since track
+      replacement is observably synchronous in the API by calling |replaceTrack|
+      multiple times. The promise exists to simplify learning when replaced
+      tracks can be stopped without impacting the recording.</div></li>
```
Pehrsons, line 462: Can you give an example of when `recorder.replaceTrack(track, withTrack); track.stop();` would impact the recording so that `track` is not replaced with `withTrack`, given that all the attributes checked synchronously are OK?

I.e., explain more in detail why the promise is needed, please.

MediaRecorder.bs
```diff
@@ -410,6 +405,67 @@ interface MediaRecorder : EventTarget {
...
+    <li>If |track| is {{ended}}, or |withTrack|'s {{kind}} differs from
+    |track|'s {{kind}}, then return a promise rejected with a newly created
+    {{InvalidModificationError}} {{DOMException}} and abort these steps.</li>
+    <li>If the <a>isolation properties</a> of |track| or |withTrack| disallow
+    access from |recorder|, return a promise rejected with a newly created
+    {{SecurityError}} {{DOMException}} and abort these steps.</li>
+    <li>If the User Agent cannot record |withTrack| in place of |track| using
+    the current configuration, then return a promise rejected with a newly
+    created {{NotSupportedError}} {{DOMException}} and abort these steps.</li>
+    <li>Let |session| be the ongoing recording session corresponding to the
```
Pehrsons, line 432: "recording session" needs a definition somewhere. There is no mention of "session" anywhere in the spec.

I can't really guess what it means either since intuitively a recorder cannot have more than one session at a time (it produces one output file across several blobs). But here it seems like the recorder has one session per track?

MediaRecorder.bs
```diff
@@ -410,6 +405,67 @@ interface MediaRecorder : EventTarget {
...
+    steps:</li>
+    <ol>
+      <li>If |recorder|'s {{state}} attribute is {{inactive}}, resolve |p|, and
+      abort these steps.</li>
+      <li>Let |replaceList| be |recorder|'s [=[[ReplaceAtOnceList]]=], and set
+      |recorder|'s [=[[ReplaceAtOnceList]]=] to an empty list.</li>
+      <li>Set |recorder|'s [=[[ReplacePromise]]=] slot to <code>null</code>, and
+      run the remaining steps in parallel:</li>
+      <li>For each recording |session| and |withTrack| pair in |replaceList|,
+      instantly switch |session| to recording |withTrack| in place of what
```
Pehrsons, line 452: s/instantly/atomically/ ?

MediaRecorder.bs
```diff
@@ -410,6 +405,67 @@ interface MediaRecorder : EventTarget {
...
+      |session| is currently recording.
```
Pehrsons, line 453: What if the track that session corresponds to has now ended?

MediaRecorder.bs
```diff
@@ -101,8 +102,8 @@ interface MediaRecorder : EventTarget {
         "False">&#10008;</span></td>
         <td class="prmOptFalse"><span role="img" aria-label=
         "False">&#10008;</span></td>
-        <td class="prmDesc">The {{MediaStream}} to be recorded. This will
-        be the value of the {{MediaRecorder/stream}} attribute.</td>
+        <td class="prmDesc">The {{MediaStream}} to be recorded. This
+        can be changed later using the {{MediaRecorder/stream}} attribute.</td>
```
Pehrsons, line 106: s/using/by setting/ ?

Is it worth noting that operations done on the stream attribute does not affect the tracks recorded (and vice versa, operations done affecting the recorded tracks does not reflect in the stream attribute's track set)?

Pehrsons (changes requested):
Three main things (more details inline),
1) I still don't see the need for the promise.
2) I don't like the complexity of this with all the new slots, but on the other hand I'm not aware of a simpler way.
3) It's worth looking into allowing the replacement of an ended track.

MediaRecorder.bs
guest271314: Why is the term "seamlessly" https://w3c.github.io/webrtc-pc/#dom-rtcrtpsender-replacetrack not used here?

> switch seamlessly to transmitting withTrack instead of the sender's existing track

MediaRecorder.bs
jan-ivar: This is consistent with other references in this spec. I'll address all of them separately in a different PR after I've moved those out of identity.

MediaRecorder.bs
jan-ivar: @henbos I believe you raised the need for a promise. Care to comment?

The only case I can think of is bad behaving JS, e.g.:
```js
recorder.replaceTrack(videoA, newVideoA);
videoA.stop();
for (const i = 0; i < 1000000; i++) x += i / 3.14;
recorder.replaceTrack(audioB, newAudioB);
audioA.stop();
```
In this case would audio and video switch at the same time?

MediaRecorder.bs
jan-ivar: It's in the note below. I based this on Henrik's https://github.com/w3c/mediacapture-record/pull/186. Happy to change it to be here if we prefer.

guest271314: Language in the body of a document is, in general, the primary source and controlling, not a note. 

https://github.com/w3c/webrtc-pc/issues/2171#issuecomment-484544660
> I don't think this note counts as citeable.

Since the opportunity is at hand would substitute "seamlessly" for "instantly" to avoid ambiguity in the text. Granted both "instantly" and "seamlessly" can also be parsed as there is no corresponding time frame referenced. If necessary substitute "SHOULD be an instant operation" for "SHOULD be a seamless operation".

guest271314: Unless "instantly" is indeed intended. Apparently that is the case given no `Promise` operation. If "instantly" is the primary term used, the term, as-applied in the specification, should be clearly defined. "seamlessly" does not infer a time component, "instantly" does, though without a reference point can be interpreted differently by any viewer or implementer.

MediaRecorder.bs
Pehrsons: Yes, but the recording would also stop immediately after they are replaced, given that the recorder is only recording `videoA` and `audioA`.

According to this PR, the first `replaceTrack` queues task A to atomically replace both `videoA` and `audioA`. `audioA.stop()` makes "all tracks become ended" (from the spec) and queues a task B to inactivate the recorder and fire "stop". Task A runs before task B, but after `audioA.stop()` so there's your answer.

Having a promise doesn't stop task B from inactivating the recorder. I still don't see the need for the promise.

I do however see a need for having task B per above exit early (the `Tracks` slot contains live tracks!), or avoid being fired (the `ReplaceAtOnceList` slot is not empty!).

MediaRecorder.bs
trackedsupport: Is there a way to add this replaceTrack method for an individual project?

ParfectShot:
any update on this MR ?

alvestrand:
Seems to have no developer interest in implementation, so closing because of lack of interest.

PR #190, opened 2019-09-26
Defer deciding codec until "onstart" fires

### LINKED ISSUE #139: Does recording of remote a/v streams always imply re-encoding?
I tried recording a remote stream from a WebRTC conference via MediaRecorder.
Even when the remote stream was h264/opus, the stream recorded locally was vp8/opus.
I am aware that I can specify a desired codec, and force encoding of h264/opus locally. But the question is weather this always involves re-encoding inside the client browser even if codecs match, or is there a way to simple dump or remux the stream, keeping the remote encoded stream without costly re-encoding?

I think this should be noted somewhere in the spec.

henbos:
Fixes #139.

This allows the User Agent to avoid the cost of re-encoding.

### DIFF
--- a/MediaRecorder.bs
+++ b/MediaRecorder.bs
@@ -276,7 +276,7 @@ interface MediaRecorder : EventTarget {
     <li>Set |recorder|'s {{state}} to {{recording}}, and run the following steps
     in parallel:
     <ol>
-      <li>The User Agent MUST decide the |recorder|'s configuration for
+      <li>Let the User Agent specify the |recorder|'s configuration for
       recording all the tracks in |tracks|, given its current constraints. If
       track data is available, the User Agent MAY take the content into account
       when making this decision as long as this does not violate how the
--- a/MediaRecorder.bs
+++ b/MediaRecorder.bs
@@ -255,9 +255,9 @@ interface MediaRecorder : EventTarget {
     <li>Let |extendedMimeType| be the value of |recorder|'s
     [=[[ConstrainedMimeType]]=] slot.</li>
 
-    <li>If |extendedMimeType| is empty, the User Agent MUST modify it by adding
-    media type, subtype and codecs parameter reflecting the configuration used
-    by the MediaRecorder to record all tracks in |tracks|, if not already
+    <li>If |extendedMimeType| is not empty, the User Agent MUST modify it by
+    adding media type, subtype and codecs parameter reflecting the configuration
+    used by the MediaRecorder to record all tracks in |tracks|, if not already
     present. This MAY include the profiles parameter [[RFC6381]] or further
     codec-specific parameters.</li>
     <div class="note">
@@ -310,11 +310,11 @@ interface MediaRecorder : EventTarget {
       </ol>
       <div class="note">
         By looking at the content of the tracks when recording start, the User
-        Agent could take frame information into account. If the MediaStreamTrack
-        is a remote track sourced from an RTCPeerConnection, the User Agent is
-        allowed to pick the same codec that the source RTP stream uses, which
-        may allow the User Agent to avoid the cost of re-encoding by recording
-        the raw encoded frames.
+        Agent could take frame information into account. For example, if the
+        MediaStreamTrack is a remote track sourced from an RTCPeerConnection,
+        the User Agent is allowed to pick the same codec that the source RTP
+        stream uses, which may allow the User Agent to avoid the cost of
+        re-encoding by recording the raw encoded frames.
       </div>
       </li>
 
--- a/MediaRecorder.bs
+++ b/MediaRecorder.bs
@@ -252,23 +252,6 @@ interface MediaRecorder : EventTarget {
     </ol>
     </li>
 
-    <li>Let |extendedMimeType| be the value of |recorder|'s
-    [=[[ConstrainedMimeType]]=] slot.</li>
-
-    <li>If |extendedMimeType| is not empty, the User Agent MUST modify it by
-    adding media type, subtype and codecs parameter reflecting the configuration
-    used by the MediaRecorder to record all tracks in |tracks|, if not already
-    present. This MAY include the profiles parameter [[RFC6381]] or further
-    codec-specific parameters.</li>
-    <div class="note">
-      In the case that |extendedMimeType| is empty, the mimeType will be updated
-      when "onstart" fires. This allows the User Agent to defer configurating
-      which codec to use until the tracks deliver frames. This allows the User
-      Agent to make optimizations based on the source of the track.
-    </div>
-
-    <li>Set |recorder|'s {{mimeType}} attribute to |extendedMimeType|.</li>
-
     <li>If |recorder|'s [=[[ConstrainedBitsPerSecond]]=] slot is not
     <code>undefined</code>, set |recorder|'s {{videoBitsPerSecond}} and
     {{audioBitsPerSecond}} attributes to values the User Agent deems reasonable
@@ -293,18 +276,18 @@ interface MediaRecorder : EventTarget {
     <li>Set |recorder|'s {{state}} to {{recording}}, and run the following steps
     in parallel:
     <ol>
-      <li>As soon as data is available on all tracks in |tracks|, if the
-      |recorder|'s current configuration does not already specify the codecs to
-      use, the User Agent MUST specify this. The User Agent MAY take the content
-      of the first frame into account when making this decision. The User Agent
-      MUST then start recording with the configuration and gather the data into
-      a {{Blob}} <var>blob</var>. As soon as recording starts, queue a task,
-      using the DOM manipulation task source, to run the following steps:
+      <li>As soon as data is available on all tracks in |tracks|, the User Agent
+      MUST specify the |recorder|'s configuration for recording the tracks, if
+      not already specified. The User Agent MAY take the content of the first
+      frame into account when making this decision. The User Agent MUST then
+      start recording with this configuration and gather the data into a
+      {{Blob}} <var>blob</var>. As soon as recording starts, queue a task, using
+      the DOM manipulation task source, to run the following steps:
       <ol>
-        <li>Set |recorder|'s {{mimeType}} to reflect the current
-        configuration, including media type, subtype and codecs parameter. This
-        MAY include the profiles parameters [[RFC6381]] or further
-        codec-specific parameters.</li>
+        <li>Set |recorder|'s {{mimeType}} to reflect the configuration,
+        including media type, subtype and codecs parameter. This MAY include the
+        profiles parameters [[RFC6381]] or further codec-specific
+        parameters.</li>
         <li><a>Fire an event</a> named <a>start</a> at <var>recorder</var>, then
         return <code>undefined</code>.</li>
       </ol>
--- a/MediaRecorder.bs
+++ b/MediaRecorder.bs
@@ -276,12 +276,11 @@ interface MediaRecorder : EventTarget {
     <li>Set |recorder|'s {{state}} to {{recording}}, and run the following steps
     in parallel:
     <ol>
-      <li>Start recording all tracks in |tracks| using the |recorder|'s current
-      configuration and gather the data into a {{Blob}} <var>blob</var>. If the
-      container and codecs to use for the recording have not yet been fully
-      specified, the User Agent specifies them in |recorder|'s current
+      <li>If the container and codecs to use for the recording have not yet been
+      fully specified, the User Agent specifies them in |recorder|'s current
       configuration. The User Agent MAY take the sources of the tracks in
-      |tracks| into account when deciding which container and codecs to use.
+      |tracks| into account when deciding which container and codecs to
+      use.
       <div class="note">
         By looking at the sources of the tracks in |tracks| when recording
         starts, the User Agent could choose a configuration that avoids
@@ -289,11 +288,13 @@ interface MediaRecorder : EventTarget {
         remote track sourced from an RTCPeerConnection, the User Agent could
         pick the same codec as is used for the MediaStreamTrack's RTP stream,
         should it be known. However, if the codec of the RTP stream changes
-        while recording the User Agent has to be prepared to re-encode it to
+        while recording, the User Agent has to be prepared to re-encode it to
         avoid interruptions.
       </div>
-      Queue a task, using the DOM manipulation task source, to run the following
-      steps:
+      </li>
+      <li>Start recording all tracks in |tracks| using the |recorder|'s current
+      configuration and gather the data into a {{Blob}} <var>blob</var>. Queue a
+      task, using the DOM manipulation task source, to run the following steps:
       <ol>
         <li>Let |extendedMimeType| be the value of |recorder|'s
         [=[[ConstrainedMimeType]]=] slot.</li>
--- a/MediaRecorder.bs
+++ b/MediaRecorder.bs
@@ -276,23 +276,24 @@ interface MediaRecorder : EventTarget {
     <li>Set |recorder|'s {{state}} to {{recording}}, and run the following steps
     in parallel:
     <ol>
-      <li>Let the User Agent specify the |recorder|'s configuration for
-      recording all the tracks in |tracks|, given its current constraints. If
-      track data is available, the User Agent MAY take the content into account
-      when making this decision as long as this does not violate how the
-      configuration has previously been constrained (see earlier steps).</li>
-      <div class="note">
-        By looking at the content of the tracks when recording start, the User
-        Agent could take frame information into account. For example, if the
-        MediaStreamTrack is a remote track sourced from an RTCPeerConnection,
-        the User Agent is allowed to pick the same codec that the source RTP
-        stream uses, which may allow the User Agent to avoid the cost of
-        re-encoding by recording the raw encoded frames.
-      </div>
       <li>Start recording all tracks in |tracks| using the |recorder|'s current
-      configuration and gather the data into a {{Blob}} <var>blob</var>. As soon
-      as recording starts, queue a task, using the DOM manipulation task source,
-      to run the following steps:
+      configuration and gather the data into a {{Blob}} <var>blob</var>. If the
+      codec or container has not yet been specified, the User Agent specifies
+      it. The User Agent MAY take the sources of the |tracks| into account when
+      making this decision as long as it does not violate the current
+      configuration.
+      <div class="note">
+        By looking at the sources of the tracks in |tracks| when recording
+        starts, the User Agent could choose a configuration that avoids
+        re-encoding track content. For example, if the MediaStreamTrack is a
+        remote track sourced from an RTCPeerConnection, the User Agent could
+        pick the same codec as is used for the MediaStreamTrack's RTP stream,
+        should it be known. However, if the codec of the RTP stream changes
+        while recording the User Agent have to be prepared to re-encode it to
+        avoid interruptions.
+      </div>
+      Queue a task, using the DOM manipulation task source, to run the following
+      steps:
       <ol>
         <li>Let |extendedMimeType| be the value of |recorder|'s
         [=[[ConstrainedMimeType]]=] slot.</li>
--- a/MediaRecorder.bs
+++ b/MediaRecorder.bs
@@ -276,21 +276,11 @@ interface MediaRecorder : EventTarget {
     <li>Set |recorder|'s {{state}} to {{recording}}, and run the following steps
     in parallel:
     <ol>
-      <li>As soon as data is available on all tracks in |tracks|, the User Agent
-      MUST specify the |recorder|'s configuration for recording the tracks, if
-      not already specified. The User Agent MAY take the content of the first
-      frame into account when making this decision. The User Agent MUST then
-      start recording with this configuration and gather the data into a
-      {{Blob}} <var>blob</var>. As soon as recording starts, queue a task, using
-      the DOM manipulation task source, to run the following steps:
-      <ol>
-        <li>Set |recorder|'s {{mimeType}} to reflect the configuration,
-        including media type, subtype and codecs parameter. This MAY include the
-        profiles parameters [[RFC6381]] or further codec-specific
-        parameters.</li>
-        <li><a>Fire an event</a> named <a>start</a> at <var>recorder</var>, then
-        return <code>undefined</code>.</li>
-      </ol>
+      <li>The User Agent MUST decide the |recorder|'s configuration for
+      recording all the tracks in |tracks|, given its current constraints. If
+      track data is available, the User Agent MAY take the content into account
+      when making this decision as long as this does not violate how the
+      configuration has previously been constrained (see earlier steps).</li>
       <div class="note">
         By looking at the content of the tracks when recording start, the User
         Agent could take frame information into account. For example, if the
@@ -299,6 +289,21 @@ interface MediaRecorder : EventTarget {
         stream uses, which may allow the User Agent to avoid the cost of
         re-encoding by recording the raw encoded frames.
       </div>
+      <li>Start recording all tracks in |tracks| using the |recorder|'s current
+      configuration and gather the data into a {{Blob}} <var>blob</var>. As soon
+      as recording starts, queue a task, using the DOM manipulation task source,
+      to run the following steps:
+      <ol>
+        <li>Let |extendedMimeType| be the value of |recorder|'s
+        [=[[ConstrainedMimeType]]=] slot.</li>
+        <li>Modify |extendedMimeType| by adding media type, subtype and codecs
+        parameter reflecting the configuration used by the MediaRecorder to
+        record all tracks in |tracks|, if not already present. This MAY include
+        the profiles parameter [[RFC6381]] or further codec-specific
+        parameters.</li>
+        <li>Set |recorder|'s {{mimeType}} attribute to |extendedMimeType|.</li>
+        <li><a>Fire an event</a> named <a>start</a> at |recorder|.</li>
+      </ol>
       </li>
 
       <li>If at any point |stream|'s <a>isolation properties</a> change so
--- a/MediaRecorder.bs
+++ b/MediaRecorder.bs
@@ -289,7 +289,7 @@ interface MediaRecorder : EventTarget {
         remote track sourced from an RTCPeerConnection, the User Agent could
         pick the same codec as is used for the MediaStreamTrack's RTP stream,
         should it be known. However, if the codec of the RTP stream changes
-        while recording the User Agent have to be prepared to re-encode it to
+        while recording the User Agent has to be prepared to re-encode it to
         avoid interruptions.
       </div>
       Queue a task, using the DOM manipulation task source, to run the following
--- a/MediaRecorder.bs
+++ b/MediaRecorder.bs
@@ -278,10 +278,10 @@ interface MediaRecorder : EventTarget {
     <ol>
       <li>Start recording all tracks in |tracks| using the |recorder|'s current
       configuration and gather the data into a {{Blob}} <var>blob</var>. If the
-      codec or container has not yet been specified, the User Agent specifies
-      it. The User Agent MAY take the sources of the |tracks| into account when
-      making this decision as long as it does not violate the current
-      configuration.
+      container and codecs to use for the recording have not yet been fully
+      specified, the User Agent specifies them in |recorder|'s current
+      configuration. The User Agent MAY take the sources of the tracks in
+      |tracks| into account when deciding which container and codecs to use.
       <div class="note">
         By looking at the sources of the tracks in |tracks| when recording
         starts, the User Agent could choose a configuration that avoids
--- a/MediaRecorder.bs
+++ b/MediaRecorder.bs
@@ -255,18 +255,16 @@ interface MediaRecorder : EventTarget {
     <li>Let |extendedMimeType| be the value of |recorder|'s
     [=[[ConstrainedMimeType]]=] slot.</li>
 
-    <li>Modify |extendedMimeType| by adding media type, subtype and codecs
-    parameter reflecting the configuration used by the MediaRecorder to record
-    all tracks in |tracks|, if not already present. This MAY include the
-    profiles parameter [[RFC6381]] or further codec-specific parameters. If
-    |extendedMimeType| did not already have a codec specified, the User Agent
-    MAY defer deciding which codec and parameters to use until recording has
-    started, in which case |extendedMimeType| will be missing this
-    information.</li>
+    <li>If |extendedMimeType| is empty, the User Agent MUST modify it by adding
+    media type, subtype and codecs parameter reflecting the configuration used
+    by the MediaRecorder to record all tracks in |tracks|, if not already
+    present. This MAY include the profiles parameter [[RFC6381]] or further
+    codec-specific parameters.</li>
     <div class="note">
-      Deferring configurating which codec to use allows the User Agent to make
-      optimizations based on the source of the track. The mimeType is guaranteed
-      to be contain codec and parameter information when "onstart" fires.
+      In the case that |extendedMimeType| is empty, the mimeType will be updated
+      when "onstart" fires. This allows the User Agent to defer configurating
+      which codec to use until the tracks deliver frames. This allows the User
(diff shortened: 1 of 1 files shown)


henbos:
Please take a look @Pehrsons and @jan-ivar

MediaRecorder.bs
```diff
@@ -286,11 +295,30 @@ interface MediaRecorder : EventTarget {
...
+        return <code>undefined</code>.</li>
+      </ol>
+      <div class="note">
+        By looking at the content of the tracks when recording start, the User
+        Agent could take frame information into account such as if the video is
+        high or low resolution. If the MediaStreamTrack is a remote track
+        sourced from an RTCPeerConnection, the User Agent is allowed to pick the
+        same codec that the source RTP stream uses, which may allow the User
+        Agent to avoid the cost of re-encoding by recording the raw encoded
+        frames.
```
henbos: I think the target audioBitsPerSecond and videoBitsPerSecond doesn't make sense if we are not re-encoding, so I would suggest allowing these to be nullable or saying that the value 0 may be used, but I would prefer to clarify this in a separate issue and PR. Here: Is this the right approach?

MediaRecorder.bs
```diff
@@ -258,7 +258,16 @@ interface MediaRecorder : EventTarget {
...
-    profiles parameter [[RFC6381]] or further codec-specific parameters.</li>
+    profiles parameter [[RFC6381]] or further codec-specific parameters. If
+    |extendedMimeType| did not already have a codec specified, the User Agent
+    MAY defer deciding which codec and parameters to use until recording has
+    started, in which case |extendedMimeType| will be missing this
+    information.</li>
+    <div class="note">
+      Deferring configurating which codec to use allows the User Agent to make
+      optimizations based on the source of the track. The mimeType is guaranteed
+      to be contain codec and parameter information when "onstart" fires.
```
alvestrand: grammarian nitpick: "guaranteed to be contain" -> "guaranteed to contain"

MediaRecorder.bs
```diff
@@ -286,11 +295,30 @@ interface MediaRecorder : EventTarget {
...
+      <ol>
+        <li>Set |recorder|'s {{mimeType}} to reflect the current
+        configuration, including codec and parameter information.</li>
+        <li><a>Fire an event</a> named <a>start</a> at <var>recorder</var>, then
+        return <code>undefined</code>.</li>
+      </ol>
+      <div class="note">
+        By looking at the content of the tracks when recording start, the User
+        Agent could take frame information into account such as if the video is
+        high or low resolution. If the MediaStreamTrack is a remote track
```
Pehrsons: Resolution is a bad example, because it can change on the fly at any time. You also don't mention what it could be used for, so I'd rather just leave it out.

MediaRecorder.bs
```diff
@@ -286,11 +295,30 @@ interface MediaRecorder : EventTarget {
...
     in parallel:
     <ol>
-      <li>Start recording all tracks in |tracks| using the {{MediaRecorder}}'s
-      current configuration, and gather the data into a {{Blob}} <var>blob</var>
-      and queue a task, using the DOM manipulation task source, to
-      <a>fire an event</a> named <a>start</a> at <var>recorder</var>, then return
-      <code>undefined</code>.</li>
+      <li>As soon as data is available on all tracks in |tracks|, if the
+      |recorder|'s current configuration does not already specify the codec and
+      parameters to use, the User Agent MUST specify this. The User Agent MAY
```
Pehrsons: I would only mention codecs here, since any other parameters (as far as mime type goes) are behind a MAY.

MediaRecorder.bs
```diff
@@ -258,7 +258,16 @@ interface MediaRecorder : EventTarget {
     <li>Modify |extendedMimeType| by adding media type, subtype and codecs
     parameter reflecting the configuration used by the MediaRecorder to record
     all tracks in |tracks|, if not already present. This MAY include the
-    profiles parameter [[RFC6381]] or further codec-specific parameters.</li>
+    profiles parameter [[RFC6381]] or further codec-specific parameters. If
+    |extendedMimeType| did not already have a codec specified, the User Agent
```
Pehrsons: Wasn't the consensus to only allow transcoding when no mimeType was given to the `MediaRecorderOptions` at all (aka the empty string)? Or did I misunderstand the TPAC minutes?

MediaRecorder.bs
```diff
@@ -258,7 +258,16 @@ interface MediaRecorder : EventTarget {
...
+    MAY defer deciding which codec and parameters to use until recording has
+    started, in which case |extendedMimeType| will be missing this
```
Pehrsons: This says the media type ("audio" or "video") and subtype (container) must be defined at this point. But if you decide on "video/webm" and a recorded track is sourced from an h264 rtp stream, it'd be illegal to stick that into webm.

You'd have to defer extending the mime type altogether. All or nothing.

MediaRecorder.bs
```diff
@@ -258,7 +258,16 @@ interface MediaRecorder : EventTarget {
...
+    information.</li>
+    <div class="note">
+      Deferring configurating which codec to use allows the User Agent to make
```
Pehrsons: s/configurating/configuring/

MediaRecorder.bs
Pehrsons: "codecs parameter" is used elsewhere since the parameter containing the codecs is called "codecs". There are other parameters too but they're optional.

Pehrsons (changes requested):
I'm thinking it'd be simpler and clearer to just move the step of `Modify |extendedMimeType| by adding media type, subtype and codecs parameter...` to when firing "start". I.e., to not even set the mimeType attribute synchronously in `start()`.

The spec change for setting it synchronously in `start()` was recent enough that no-one is implementing it yet. I have implemented it in Firefox but those patches are still under review.

It would allow you to avoid duplicating language about extending and setting the mime type according to the configuration of the recorder.
And most of all it avoids forcing applications to handle two paths (mimetype present vs not present) where one is so rare (mimetype not preset) it's going to be easily missed.

MediaRecorder.bs
henbos: That's correct, I think there was some confusion on my part on the connection between "container" and "codec/parameters", since this string is used to specify both things. I updated the PR to match "no mimeType" per-TPAC decision, but I think we might want to have a follow-up discussion/issue about this.

MediaRecorder.bs
henbos: I think if the application asks for "video/webm" but doesn't specify the codec, it would be reasonable to allow the browser to choose codec for you as long as it is compatible with "video/webm". So for example, if the receiving stream is VP8 the browser could avoid transcoding, but if the stream is H264 then the UA would be forced to transcode in order order to fit the container. But I think this can be a separate discussion. Currently: All or nothing.

MediaRecorder.bs
henbos: OK!

MediaRecorder.bs
henbos: OK I removed it. The idea was that if the UA might be able to do a trade-off between encoders based on the resolution of the first frame (heuristic: "if the first frame is HD, this is probably an HD track, in which case encoder X is better than encoder Y"), but I don't need to get in to that, it was just something I thought of, and not the real motivation behind this note.

henbos (commented):
Thanks for the review, please take another look.

henbos:
> I'm thinking it'd be simpler and clearer to just move the step of Modify |extendedMimeType| by adding media type, subtype and codecs parameter... to when firing "start". I.e., to not even set the mimeType attribute synchronously in start().
> 
> The spec change for setting it synchronously in start() was recent enough that no-one is implementing it yet. I have implemented it in Firefox but those patches are still under review.
> 
> It would allow you to avoid duplicating language about extending and setting the mime type according to the configuration of the recorder.
> And most of all it avoids forcing applications to handle two paths (mimetype present vs not present) where one is so rare (mimetype not preset) it's going to be easily missed.

Oh that's great, it's what I would have preferred so that mimeType is always only updated in one place, that's less surprises for developers. I'll update the PR again.

henbos (commented):
Now mimeType is only updated at "onstart", PTAL :)

MediaRecorder.bs
```diff
@@ -286,11 +276,30 @@ interface MediaRecorder : EventTarget {
...
+      not already specified. The User Agent MAY take the content of the first
+      frame into account when making this decision. The User Agent MUST then
+      start recording with this configuration and gather the data into a
+      {{Blob}} <var>blob</var>. As soon as recording starts, queue a task, using
+      the DOM manipulation task source, to run the following steps:
+      <ol>
+        <li>Set |recorder|'s {{mimeType}} to reflect the configuration,
+        including media type, subtype and codecs parameter. This MAY include the
+        profiles parameters [[RFC6381]] or further codec-specific
+        parameters.</li>
```
Pehrsons, line 305: Please keep the old language and algorithm for updating the mimeType. It was carefully worked out, I'd hate to see it go.

MediaRecorder.bs
```diff
@@ -286,11 +276,30 @@ interface MediaRecorder : EventTarget {
     <li>Set |recorder|'s {{state}} to {{recording}}, and run the following steps
     in parallel:
     <ol>
-      <li>Start recording all tracks in |tracks| using the {{MediaRecorder}}'s
-      current configuration, and gather the data into a {{Blob}} <var>blob</var>
-      and queue a task, using the DOM manipulation task source, to
-      <a>fire an event</a> named <a>start</a> at <var>recorder</var>, then return
-      <code>undefined</code>.</li>
+      <li>As soon as data is available on all tracks in |tracks|, the User Agent
```
Pehrsons: This is weird because `As soon as data is available...` is not a step you can run. It's not something you introduced, it's just this spec's legacy.

Perhaps we could try to improve this at the same time, maybe by making it the below?
```
Set |recorder|'s {{state}} to {{recording}}, and as soon as data is available
on all tracks in |tracks|, run the following steps in parallel:
- Let |extendedMimeType| be the value of |recorder|'s
  [=[[ConstrainedMimeType]]=] slot.
- Modify |extendedMimeType| by adding media type, subtype and codecs
  parameter reflecting the configuration used by the MediaRecorder to record
  all tracks in |tracks|, if not already present. This MAY include the
  profiles parameter [[RFC6381]] or further codec-specific parameters.
- Set |recorder|'s {{mimeType}} attribute to |extendedMimeType|.
- Start gathering data into a {{Blob}} <var>blob</var>.
- <a>Fire an event</a> named <a>start</a> at |recorder|.
```

Followed by your note.

An issue I can think of however is: if one track doesn't have any data for a long time -- what happens to the data on other tracks? As this stands it will be discarded, but that would regress current MediaRecorder behavior which records it.

Could we solve that? Some sort of track-by-track approach perhaps?

@jan-ivar might have thoughts too, he's much more seasoned in spec writing than myself.

Pehrsons (changes requested):

MediaRecorder.bs
henbos: > An issue I can think of however is: if one track doesn't have any data for a long time -- what happens to the data on other tracks? As this stands it will be discarded, but that would regress current MediaRecorder behavior which records it.
> 
> Could we solve that? Some sort of track-by-track approach perhaps?

This does change the spec to say not to record until you have data, which could cause a regression, however:
- Based on a brief chat with someone, I don't _think_ that current implementations actually start recording prior to having data - though note that I have not verified if this is true.
- I seriously doubt that applications are relying on recording starting even though their tracks are muted at the time of starting. What's the use case of this?

I prefer not making the API more complicated by having track-by-track basis of starting recording. Just like it is best to move updating mimeType to "onstart", I think it's best to move the start of recording to when we know how we want to record, even if that is a change in behavior.

henbos: I like your other suggestions, will update PR when I have time!

Pehrsons: > > An issue I can think of however is: if one track doesn't have any data for a long time -- what happens to the data on other tracks? As this stands it will be discarded, but that would regress current MediaRecorder behavior which records it.
> > Could we solve that? Some sort of track-by-track approach perhaps?
> 
> This does change the spec to say not to record until you have data, which could cause a regression, however:
> * Based on a brief chat with someone, I don't _think_ that current implementations actually start recording prior to having data - though note that I have not verified if this is true.

Firefox will start as soon as all tracks are available. And with current WIP patches I have (making us compliant with the current spec), this means it will start synchronously in `start()`. I.e., it doesn't wait for data on all tracks -- any track that doesn't have data initially will get padding or a delayed start time as needed. It does wait for output from the muxer (i.e., it waits for data to get written to the blob) before firing "start".

> * I seriously doubt that applications are relying on recording starting even though their tracks are muted at the time of starting. What's the use case of this?

One use case it could allow in the future would be to let `Stop();Start();` if they happen in the same task, stop the current recording and start a new one in a seamless way. The full blobs produced by two such recordings could be played seamlessly with MSE. It would be one way to handle track-set changes on the fly.

With your proposed changes we lock ourselves out of that possibility.

> I prefer not making the API more complicated by having track-by-track basis of starting recording. Just like it is best to move updating mimeType to "onstart", I think it's best to move the start of recording to when we know how we want to record, even if that is a change in behavior.

henbos: > One use case it could allow in the future would be to let Stop();Start(); if they happen in the same task, stop the current recording and start a new one in a seamless way. The full blobs produced by two such recordings could be played seamlessly with MSE. It would be one way to handle track-set changes on the fly.
> 
> With your proposed changes we lock ourselves out of that possibility.

I'm happy to lock ourselves out of that possibility. I don't think we should make an operation like "replace track" be implemented by a series of method calls relying on being invoked in the same task, it seems a bit implicit, affecting implementation strategies and possibly limiting or adding complexity, and API ergonomics-wise, a step like "stop" in the middle not being descriptive of what we're trying to achieve. Besides, if we achieve "replace track" by chaining two existing methods with existing uses, how do you feature detect for "replace track"?

Anyway that's my personal opinion. But at TPAC Fukuoka 2019 the W3C working group [consensus was reached](https://www.w3.org/2019/09/18-webrtc-minutes.html#item09) to add a replaceTrack() method (slides [76](https://docs.google.com/presentation/d/1Xprz0ElNTUWRDdX6kyp11cZ4ItTvUFUmzoXsKZdXxuU/edit#slide=id.g62a86e6d8c_3_12)-[77](https://docs.google.com/presentation/d/1Xprz0ElNTUWRDdX6kyp11cZ4ItTvUFUmzoXsKZdXxuU/edit#slide=id.g634b2fa483_43_0)) to handle that use case explicitly, so I think we're free to do what we want here.

jan-ivar: I think the [resolution](https://www.w3.org/2019/09/18-webrtc-minutes.html#resolution09) was to *"Start with proposal B, Jan-Ivar to create a PR, promise based."* - I read that as there being interest in pursuing `replaceTrack`. We don't have a final PR to have consensus over.

Regardless, I think they solve different problems:
 1. `replaceTrack` seamlessly merges two sources into one container.
 2. `rec.stop(); rec.start()` splits one source into two (independently playable) containers.

Pehrsons: > > One use case it could allow in the future would be to let Stop();Start(); if they happen in the same task, stop the current recording and start a new one in a seamless way. The full blobs produced by two such recordings could be played seamlessly with MSE. It would be one way to handle track-set changes on the fly.
> > With your proposed changes we lock ourselves out of that possibility.
> 
> I'm happy to lock ourselves out of that possibility. I don't think we should make an operation like "replace track" be implemented by a series of method calls relying on being invoked in the same task

It wouldn't be "replace track". It would most likely be "add track" since that's the only thing that cannot be fulfilled by the current spec or the `replaceTrack` proposal.

> , it seems a bit implicit, affecting implementation strategies and possibly limiting or adding complexity, and API ergonomics-wise, a step like "stop" in the middle not being descriptive of what we're trying to achieve. Besides, if we achieve "replace track" by chaining two existing methods with existing uses, how do you feature detect for "replace track"?
> 
> Anyway that's my personal opinion. But at TPAC Fukuoka 2019 the W3C working group [consensus was reached](https://www.w3.org/2019/09/18-webrtc-minutes.html#item09) to add a replaceTrack() method (slides [76](https://docs.google.com/presentation/d/1Xprz0ElNTUWRDdX6kyp11cZ4ItTvUFUmzoXsKZdXxuU/edit#slide=id.g62a86e6d8c_3_12)-[77](https://docs.google.com/presentation/d/1Xprz0ElNTUWRDdX6kyp11cZ4ItTvUFUmzoXsKZdXxuU/edit#slide=id.g634b2fa483_43_0)) to handle that use case explicitly, so I think we're free to do what we want here.

Ok. If we can't solve this I'd rather see us skip the mimetype deferral and only add a note. The application can ensure the track has data with the "mute" event like you [mentioned before](https://github.com/w3c/mediacapture-record/issues/139#issuecomment-535480592), should it want the UA to be able to take the received-track's codec into account. It could of course also take a guess and hope that they match.

henbos (commented):
Sorry for the delay I've been short on time. Please take another look.
- I reused the same language as prior to this PR as suggested. I moved some things slightly from the way you phrased it so that recording starts when we are "in parallel" (like prior to this PR) and that the internal slots are only touched in the "queue a task" (so that we don't introduce touching internal slots off-thread). The language I'm using for specifying the configuration "given its current constraints" is meant to refer to the existing language that in start() says "Constrain the configuration".
- I removed the waiting for data the become available.

As such, the only intended observable change in this PR is that mimeType is updated at "onstart" rather than "start()".

henbos:
> Ok. If we can't solve this I'd rather see us skip the mimetype deferral and only add a note.

Let's not throw everything out over a difference of opinion about whether or not to wait for data to become available, I'm happy to not make that change, as long as we don't force the UA to set the mimeType prior to starting the recording. See PR, I hope this will make both sides happy?

MediaRecorder.bs
```diff
@@ -286,11 +276,35 @@ interface MediaRecorder : EventTarget {
...
-      <li>Start recording all tracks in |tracks| using the {{MediaRecorder}}'s
-      current configuration, and gather the data into a {{Blob}} <var>blob</var>
-      and queue a task, using the DOM manipulation task source, to
-      <a>fire an event</a> named <a>start</a> at <var>recorder</var>, then return
-      <code>undefined</code>.</li>
+      <li>Let the User Agent specify the |recorder|'s configuration for
+      recording all the tracks in |tracks|, given its current constraints. If
+      track data is available, the User Agent MAY take the content into account
+      when making this decision as long as this does not violate how the
+      configuration has previously been constrained (see earlier steps).</li>
```
henbos: I just realized: If we want re-encoding to be allowed to be avoided "_only when mimeType is unset by JS_" as we decided at TPAC, I might need to add a sentence about not being allowed to avoid re-encoding is [[ConstrainedMimeType]] is non-null.

As-is I think the steps are correct for always deferring mimeType which I think we want, but the note makes it sound like we are allowed to avoid re-encoding as long as we don't violate the constraints of [[ConstrainedMimeType]], but if we want explicitly asking for VP8 to cause re-encoding then we should clarify that.

But @Pehrsons please take a look the way it is written right now anyway to make sure we're on the right track.

Pehrsons:
> > Ok. If we can't solve this I'd rather see us skip the mimetype deferral and only add a note.
> 
> Let's not throw everything out over a difference of opinion about whether or not to wait for data to become available, I'm happy to not make that change, as long as we don't force the UA to set the mimeType prior to starting the recording. See PR, I hope this will make both sides happy?

Primarily I want to avoid a situation where we set the mimeType at different times depending on the UA's intentions (sync in start() if transcoding, vs async when start() later fires "start" when avoiding transcoding). Your PR achieves that.

Secondarily I want to avoid shutting the door on something that would allow an application to dynamically handle track set changes (FWIW issue #4 covered this and is one of the most discussed issues on the spec, but was closed without resolution). Your PR does not achieve that. I could accept shutting this door if the WG discusses this and finds consensus for shutting it.

I see two ways out that avoid both situations above:
- Letting the UA start encoding data synchronously in start() although it won't fire "start" (or reveal its mimeType) yet.
- Letting the UA guess (I imagine there's a bunch of heuristics the UA could use to guess right) mimeType synchronously in start() and avoiding transcoding iff the guessed codec and the actual codec match.

I'll review the details of the PR if we can move forward with one of the three options mentioned above.

henbos:
> Secondarily I want to avoid shutting the door on something that would allow an application to dynamically handle track set changes (FWIW issue #4 covered this and is one of the most discussed issues on the spec, but was closed without resolution). Your PR does not achieve that. I could accept shutting this door if the WG discusses this and finds consensus for shutting it.

My PR seems largely irrelevant from that issue.
Prior to my PR it said "Decide mimeType synchronously, then in parallel start encoding."
After my PR it says "In parallel, decide mimeType and start encoding."

I have not changed when encoding starts. The only change to be spec-compliant before vs now is to not surface the mimeType synchronously. If anything this adds flexibility, it doesn't remove it. If you think this is a problem, please explain what you mean.

> I see two ways out that avoid both situations above:
> 
> - Letting the UA start encoding data synchronously in start() although it won't fire "start" (or reveal its mimeType) yet.

If you want to change when encoding starts I think that should be discussed in a separate issue and PR. This is already something that happens in parallel. And considering encoding is something that happens in parallel, I don't think it makes sense to say to "start encoding synchronously" unless you intend to move encoding to the main thread. The best you can get is "in parallel, start encoding as soon as possible". If you want to keep track of order of starting and stopping and being able to control things synchronously for the purpose of modifying track sets I think you can do that _with or without this PR_. Again, that seems like a separate issue?

> - Letting the UA guess (I imagine there's a bunch of heuristics the UA could use to guess right) mimeType synchronously in start() and avoiding transcoding iff the guessed codec and the actual codec match.
> 
> I'll review the details of the PR if we can move forward with one of the three options mentioned above.

Why does the application - or implementation for that matter - need to know mimeType synchronously at start() when encoding, and all events related to encoding (onstart, onstop, ondataavailable, onpause, onresume, onerror) happen later?

henbos:
@jan-ivar Do you have input?

MediaRecorder.bs
```diff
@@ -286,11 +276,35 @@ interface MediaRecorder : EventTarget {
...
+      <ol>
+        <li>Let |extendedMimeType| be the value of |recorder|'s
+        [=[[ConstrainedMimeType]]=] slot.</li>
+        <li>Modify |extendedMimeType| by adding media type, subtype and codecs
+        parameter reflecting the configuration used by the MediaRecorder to
+        record all tracks in |tracks|, if not already present. This MAY include
+        the profiles parameter [[RFC6381]] or further codec-specific
+        parameters.</li>
+        <li>Set |recorder|'s {{mimeType}} attribute to |extendedMimeType|.</li>
+        <li><a>Fire an event</a> named <a>start</a> at |recorder|.</li>
```
alvestrand, line 307: This language as written allows the MIME type to be modified "as soon as recording starts", not just to be "set if unset". Is that intentional?

MediaRecorder.bs
```diff
@@ -286,11 +276,35 @@ interface MediaRecorder : EventTarget {
...
+      <div class="note">
+        By looking at the content of the tracks when recording start, the User
+        Agent could take frame information into account. For example, if the
+        MediaStreamTrack is a remote track sourced from an RTCPeerConnection,
+        the User Agent is allowed to pick the same codec that the source RTP
+        stream uses, which may allow the User Agent to avoid the cost of
+        re-encoding by recording the raw encoded frames.
+      </div>
+      <li>Start recording all tracks in |tracks| using the |recorder|'s current
+      configuration and gather the data into a {{Blob}} <var>blob</var>. As soon
```
alvestrand: Formatting: At "As soon as", are you still in the same execution cycle?
If "recording starts" doesn't happen in the same execution cycle, it would be clearer to take this step and what follows out of this numbered list and start a new one.

MediaRecorder.bs
```diff
@@ -286,11 +276,35 @@ interface MediaRecorder : EventTarget {
     <li>Set |recorder|'s {{state}} to {{recording}}, and run the following steps
     in parallel:
     <ol>
-      <li>Start recording all tracks in |tracks| using the {{MediaRecorder}}'s
-      current configuration, and gather the data into a {{Blob}} <var>blob</var>
-      and queue a task, using the DOM manipulation task source, to
-      <a>fire an event</a> named <a>start</a> at <var>recorder</var>, then return
-      <code>undefined</code>.</li>
+      <li>Let the User Agent specify the |recorder|'s configuration for
+      recording all the tracks in |tracks|, given its current constraints. If
```
Pehrsons: There's already a current configuration that has been constrained above. "current constraints" would be a new concept, and "specify the recorder's configuration" seems to overwrite the current configuration.

Instead of this paragraph, perhaps "Start recording all tracks in |tracks| using the recorder's current configuration" can be amended to mention that there might be room to choose container and codec still?

MediaRecorder.bs
Pehrsons: Do we want to limit any container/codec choice heuristic to only use the constrained-mime, the track set (from above) and track data? What if we're recording a single track that is sourced from a peer connection that has not yet received any track data, yet has finished negotiating a single codec? By this language we'd have to ignore that codec already chosen by the peer connection. Similarly for a captured media element that has metadata but hasn't started playing yet.

MediaRecorder.bs
```diff
@@ -286,11 +276,35 @@ interface MediaRecorder : EventTarget {
...
+      track data is available, the User Agent MAY take the content into account
+      when making this decision as long as this does not violate how the
+      configuration has previously been constrained (see earlier steps).</li>
+      <div class="note">
+        By looking at the content of the tracks when recording start, the User
+        Agent could take frame information into account. For example, if the
+        MediaStreamTrack is a remote track sourced from an RTCPeerConnection,
+        the User Agent is allowed to pick the same codec that the source RTP
+        stream uses, which may allow the User Agent to avoid the cost of
+        re-encoding by recording the raw encoded frames.
```
Pehrsons: I think this needs rephrasing because the encoding of the rtp stream is not inherently part of "frame information".

Maybe "By looking at the sources of the tracks in |tracks| when recording starts, the User Agent could choose a configuration that avoids re-encoding track content. For example, if the MediaStreamTrack is a remote track sourced from an RTCPeerConnection, the User Agent could pick the same codec as is used for the MediaStreamTrack's RTP stream, should it be known."?

And perhaps mention that the RTP stream codec could change on the fly so the MediaRecorder must at least be able to encode it itself, to avoid an interrupted recording. Choosing a codec like that seems forbidden by the "For each track in |tracks|, if the User Agent cannot record the track using the current configuration, then throw a {{NotSupportedError}} {{DOMException}} and abort all steps." above, but it's probably good to re-iterate on that.

Pehrsons:
@henbos I must have misread your previous comment. Apologies.

This approach looks fine. I'll see if I have any detailed comments to put in a review.

Pehrsons (changes requested):

MediaRecorder.bs
henbos: I rephrased it, does this work better?

MediaRecorder.bs
henbos: I like your way of phrasing this better than mine and by saying "sources" we don't close any doors unnecessarily (e.g. only a single codec being negotiated).

I also added a sentence about codec changing on the fly. Our implementer would probably like to fire "onerror" if the codec changes on the fly, but I think a change like that would require a separate issue and discussions. For now it says "if the codec of the RTP stream changes while recording the User Agent have to be prepared to re-encode it to avoid interruptions" so that UAs will produce a recording for the entire stream whether or not re-enocoding was ever avoided.

MediaRecorder.bs
henbos: We're already in "in parallel" steps, so queuing a task will happen in the next execution cycle. The "as soon as" may be superflous given that we've already said to start the encoding but I wanted to be clear that "onstart" happens as a result of starting to encode. But I removed the "as soon as" in case that was confusing.

MediaRecorder.bs
henbos: Yes mimeType is modified to match the configuration. This is the same text as prior to this PR, except it was moved to "onstart" instead of happening prior to starting encoding at start().

henbos (commented):
PTAL :)

MediaRecorder.bs
```diff
@@ -286,11 +276,36 @@ interface MediaRecorder : EventTarget {
...
-      current configuration, and gather the data into a {{Blob}} <var>blob</var>
-      and queue a task, using the DOM manipulation task source, to
-      <a>fire an event</a> named <a>start</a> at <var>recorder</var>, then return
-      <code>undefined</code>.</li>
+      <li>Start recording all tracks in |tracks| using the |recorder|'s current
+      configuration and gather the data into a {{Blob}} <var>blob</var>. If the
+      codec or container has not yet been specified, the User Agent specifies
+      it. The User Agent MAY take the sources of the |tracks| into account when
+      making this decision as long as it does not violate the current
+      configuration.
```
Pehrsons: I'd make it "container or codec" since an application may constrain a container without constraining a codec, but not the other way around.

The are multiple codecs per recording, so "codec" in your text needs to be plural. There's at least one audio codec and one video codec when the MediaRecorder is recording such tracks.

I'd maybe make some other small changes like:
"If the container and codecs to use for the recording have not yet been fully specified, the User Agent specifies them in |recorder|'s current configuration. The User Agent MAY take the sources of the tracks in |tracks| into account when deciding which container and codecs to use."
This is mostly by personal preference, do with it as you wish.

Pehrsons (approved):
LGTM with nits, thanks for your patience!

I'll summon jib for a final look with his editor's hat (though another spec admittedly) on.

MediaRecorder.bs
henbos: I updated the text to use your suggestion.

henbos (commented):
> LGTM with nits, thanks for your patience!
> 
> I'll summon jib for a final look with his editor's hat (though another spec admittedly) on.

Thank you for your patience as well :)

PTAL @jan-ivar, me and @Pehrsons are happy now.

MediaRecorder.bs
```diff
@@ -286,11 +276,36 @@ interface MediaRecorder : EventTarget {
...
-      current configuration, and gather the data into a {{Blob}} <var>blob</var>
-      and queue a task, using the DOM manipulation task source, to
-      <a>fire an event</a> named <a>start</a> at <var>recorder</var>, then return
-      <code>undefined</code>.</li>
+      <li>Start recording all tracks in |tracks| using the |recorder|'s current
+      configuration and gather the data into a {{Blob}} <var>blob</var>. If the
+      container and codecs to use for the recording have not yet been fully
+      specified, the User Agent specifies them in |recorder|'s current
+      configuration. The User Agent MAY take the sources of the tracks in
+      |tracks| into account when deciding which container and codecs to use.
```
jan-ivar: This seems out of order to me, using "current configuration" before modifying it. Also, I worry the scope of the MAY here is unclear. How about:
 1. If the container and codecs to use for the recording have not yet been fully specified, specify them in |recorder|'s current configuration. The User Agent MAY take the sources of the tracks in |tracks| into account when making this decision.
2. Start recording all tracks in |tracks| using the |recorder|'s current configuration, and gather the data into a {{Blob}} <var>blob</var>.

MediaRecorder.bs
```diff
@@ -286,11 +276,36 @@ interface MediaRecorder : EventTarget {
...
+      <div class="note">
+        By looking at the sources of the tracks in |tracks| when recording
+        starts, the User Agent could choose a configuration that avoids
+        re-encoding track content. For example, if the MediaStreamTrack is a
+        remote track sourced from an RTCPeerConnection, the User Agent could
+        pick the same codec as is used for the MediaStreamTrack's RTP stream,
+        should it be known. However, if the codec of the RTP stream changes
+        while recording the User Agent has to be prepared to re-encode it to
```
jan-ivar: comma after recording?

jan-ivar (approved):
Lgtm, just editorial nits.

henbos (commented):
Done! :)

Pehrsons:
I'm updating WPT in https://bugzilla.mozilla.org/show_bug.cgi?id=1590997. See https://github.com/web-platform-tests/wpt/pull/20171.

PR #216, opened 2023-01-12
Add keyframe interval support.

handellm:
This PR adds support for video key frame configurability in the MediaRecorder API.

***
<a href="https://pr-preview.s3.amazonaws.com/handellm/mediacapture-record/pull/216.html" title="Last updated on Apr 20, 2023, 2:08 PM UTC (f6c76cb)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/mediacapture-record/216/c71f537...handellm:f6c76cb.html" title="Last updated on Apr 20, 2023, 2:08 PM UTC (f6c76cb)">Diff</a>

### DIFF
--- a/MediaRecorder.bs
+++ b/MediaRecorder.bs
@@ -99,15 +99,15 @@ interface MediaRecorder : EventTarget {
     <li>Let |recorder| have a <dfn>\[[ConstrainedBitsPerSecond]]</dfn> internal
     slot, initialized to the value of |options|'
     {{MediaRecorderOptions/bitsPerSecond}} member if it is present, otherwise
-    <code>undefined</code>.</li>
+    <code>null</code>.</li>
     <li>Let |recorder| have a <dfn>\[[VideoKeyFrameIntervalDuration]]</dfn> internal
     slot, initialized to the value of |options|'
     {{MediaRecorderOptions/videoKeyFrameIntervalDuration}} member if it is present, otherwise
-    <code>undefined</code>.</li>
+    <code>null</code>.</li>
     <li>Let |recorder| have a <dfn>\[[VideoKeyFrameIntervalCount]]</dfn> internal
     slot, initialized to the value of |options|'
     {{MediaRecorderOptions/videoKeyFrameIntervalCount}} member if it is present, otherwise
-    <code>undefined</code>.</li>
+    <code>null</code>.</li>
     <li>Initialize |recorder|'s {{MediaRecorder/stream}} attribute to
     |stream|.</li>
     <li>Initialize |recorder|'s {{MediaRecorder/mimeType}} attribute to the
--- a/MediaRecorder.bs
+++ b/MediaRecorder.bs
@@ -247,11 +247,13 @@ interface MediaRecorder : EventTarget {
 
     <li>Let |videoKeyFrameInterval| be the value of |recorder|'s internal
     [=[[VideoKeyFrameInterval]]=] slot if defined, and configure the video encoder to
-    target key frame generation with the interval. If the value isn't defined, the
-    User Agent will emit key frames as it deems fit for the purpose.
+    target key frame generation with the interval, meaning the User Agent SHOULD
+    encode a key frame on the next frame after this interval has passed since the
+    last key frame. If |videoKeyFrameInterval| isn't defined, the User Agent will
+    emit key frames as it deems fit.
     |videoKeyFrameInterval| is a hint for the video encoder and the actual
     intervals in the recording might be surpassed or not achieved. Note also that
-    encoders might make independent decisions when to emit key frames.</li>
+    encoders will make independent decisions about when to emit key frames.</li>
 
     <li>Constrain the configuration of |recorder| to encode using the
     {{BitrateMode}} specified by the value of |recorder|'s {{MediaRecorder/audioBitrateMode}}
--- a/MediaRecorder.bs
+++ b/MediaRecorder.bs
@@ -98,11 +98,15 @@ interface MediaRecorder : EventTarget {
     member.</li>
     <li>Let |recorder| have a <dfn>\[[ConstrainedBitsPerSecond]]</dfn> internal
     slot, initialized to the value of |options|'
-    {{MediaRecorderOptions/bitsPerSecond}} member, if it is present, otherwise
+    {{MediaRecorderOptions/bitsPerSecond}} member if it is present, otherwise
     <code>undefined</code>.</li>
-    <li>Let |recorder| have a <dfn>\[[VideoKeyFrameInterval]]</dfn> internal
+    <li>Let |recorder| have a <dfn>\[[VideoKeyFrameIntervalDuration]]</dfn> internal
     slot, initialized to the value of |options|'
-    {{MediaRecorderOptions/videoKeyFrameInterval}} member, if it is present, otherwise
+    {{MediaRecorderOptions/videoKeyFrameIntervalDuration}} member if it is present, otherwise
+    <code>undefined</code>.</li>
+    <li>Let |recorder| have a <dfn>\[[VideoKeyFrameIntervalCount]]</dfn> internal
+    slot, initialized to the value of |options|'
+    {{MediaRecorderOptions/videoKeyFrameIntervalCount}} member if it is present, otherwise
     <code>undefined</code>.</li>
     <li>Initialize |recorder|'s {{MediaRecorder/stream}} attribute to
     |stream|.</li>
@@ -245,15 +249,24 @@ interface MediaRecorder : EventTarget {
     the value might be surpassed, not achieved, or only be achieved over a long
     period of time.</li>
 
-    <li>Let |videoKeyFrameInterval| be the value of |recorder|'s internal
-    [=[[VideoKeyFrameInterval]]=] slot if defined, and configure the video encoder to
-    target key frame generation with the interval, meaning the User Agent SHOULD
-    encode a key frame on the next frame after this interval has passed since the
-    last key frame. If |videoKeyFrameInterval| isn't defined, the User Agent will
-    emit key frames as it deems fit.
-    |videoKeyFrameInterval| is a hint for the video encoder and the actual
-    intervals in the recording might be surpassed or not achieved. Note also that
-    encoders will make independent decisions about when to emit key frames.</li>
+    <li>Let |videoKeyFrameIntervalDuration| be the value of |recorder|'s internal
+    [=[[VideoKeyFrameIntervalDuration]]=] slot if defined. Additionally let
+    |videoKeyFrameIntervalCount| be the value of |recorder|'s internal
+    [=[[VideoKeyFrameIntervalCount]]=] slot if defined.
+    If neither of the slots are defined, the User Agent may emit key frames as
+    it deems fit.
+    If only |videoKeyFrameIntervalDuration| is defined, the video encoder
+    SHOULD produce a keyframe on the first frame arriving after
+    |videoKeyFrameIntervalDuration| elapsed since the last key frame.
+    If only |videoKeyFrameIntervalCount| is defined, the video encoder
+    SHOULD produce a keyframe on the first frame arriving after
+    |videoKeyFrameIntervalCount| frames passed since the last key frame.
+    If both are defined, the video encoder SHOULD produce a keyframe on the
+    first frame arriving after both |videoKeyFrameIntervalDuration| elapsed AND
+    |videoKeyFrameIntervalCount| frames passed since the last key frame.
+
+    Note that encoders will sometimes make independent decisions about when to
+    emit key frames.</li>
 
     <li>Constrain the configuration of |recorder| to encode using the
     {{BitrateMode}} specified by the value of |recorder|'s {{MediaRecorder/audioBitrateMode}}
@@ -545,7 +558,8 @@ dictionary MediaRecorderOptions {
   unsigned long videoBitsPerSecond;
   unsigned long bitsPerSecond;
   BitrateMode audioBitrateMode = "variable";
-  DOMHighResTimeStamp videoKeyFrameInterval;
+  DOMHighResTimeStamp videoKeyFrameIntervalDuration;
+  unsigned long videoKeyFrameIntervalCount;
 };
 </pre>
 
@@ -580,9 +594,15 @@ dictionary MediaRecorderOptions {
   <dt><dfn dict-member for="MediaRecorderOptions"><code>audioBitrateMode</code></dfn></dt>
   <dd>Specifes the {{BitrateMode}} that should be used to encode the Audio track(s).</dd>
 
-  <dt><dfn dict-member for="MediaRecorderOptions"><code>videoKeyFrameInterval</code></dfn></dt>
-  <dd>Specifies the nominal interval between key frames in the encoded video stream.
-  If this value isn't specified, the UA will control key frame generation as it sees fit.</dd>
+  <dt><dfn dict-member for="MediaRecorderOptions"><code>videoKeyFrameIntervalDuration</code></dfn></dt>
+  <dd>Specifies the nominal interval in time between key frames in the encoded video stream.
+  The UA controls key frame generation considering this dictionary member as well as 
+  {{MediaRecorderOptions/videoKeyFrameIntervalCount}}.</dd>
+
+  <dt><dfn dict-member for="MediaRecorderOptions"><code>videoKeyFrameIntervalCount</code></dfn></dt>
+  <dd>Specifies the interval in number of frames between key frames in the encoded video stream.
+  The UA controls key frame generation considering this dictionary member as well as 
+  {{MediaRecorderOptions/videoKeyFrameIntervalDuration}}.</dd>
 </dl>
 
 ## BitrateMode ## {#bitratemode}
--- a/MediaRecorder.bs
+++ b/MediaRecorder.bs
@@ -249,10 +249,10 @@ interface MediaRecorder : EventTarget {
     the value might be surpassed, not achieved, or only be achieved over a long
     period of time.</li>
 
-    <li>Let |videoKeyFrameIntervalDuration| be the value of |recorder|'s internal
-    [=[[VideoKeyFrameIntervalDuration]]=] slot if defined, and let
-    |videoKeyFrameIntervalCount| be the value of |recorder|'s internal
-    [=[[VideoKeyFrameIntervalCount]]=] slot if defined. The UA SHOULD constrain the
+    <li>Let |videoKeyFrameIntervalDuration| be
+    |recorder|.[=[[VideoKeyFrameIntervalDuration]]=], and let
+    |videoKeyFrameIntervalCount| be
+    |recorder|.[=[[VideoKeyFrameIntervalCount]]=]. The UA SHOULD constrain the
     configuration of |recorder| so that the video encoders follow the below rules:
 
     <ol>
--- a/MediaRecorder.bs
+++ b/MediaRecorder.bs
@@ -123,7 +123,7 @@ interface MediaRecorder : EventTarget {
     present. Otherwise, choose a target value the User Agent deems reasonable
     for audio.</li>
     <li>If |recorder|'s [=[[ConstrainedBitsPerSecond]]=] slot is not
-    <code>undefined</code>, set |recorder|'s {{MediaRecorder/videoBitsPerSecond}} and
+    <code>null</code>, set |recorder|'s {{MediaRecorder/videoBitsPerSecond}} and
     {{MediaRecorder/audioBitsPerSecond}} attributes to values the User Agent deems reasonable
     for the respective media types, such that the sum of {{MediaRecorder/videoBitsPerSecond}}
     and {{MediaRecorder/audioBitsPerSecond}} is close to the value of |recorder|'s
@@ -228,8 +228,8 @@ interface MediaRecorder : EventTarget {
     media type, container, and codec specified in the [=[[ConstrainedMimeType]]=]
     slot.</li>
 
-    <li>If |recorder|'s [=[[ConstrainedBitsPerSecond]]=] slot is not
-    <code>undefined</code>, set |recorder|'s {{MediaRecorder/videoBitsPerSecond}} and
+    <li>If |recorder|'s [=[[ConstrainedBitsPerSecond]]=] slot is
+    <code>null</code>, set |recorder|'s {{MediaRecorder/videoBitsPerSecond}} and
     {{MediaRecorder/audioBitsPerSecond}} attributes to values the User Agent deems reasonable
     for the respective media types, for recording all tracks in |tracks|, such
     that the sum of {{MediaRecorder/videoBitsPerSecond}} and {{MediaRecorder/audioBitsPerSecond}} is close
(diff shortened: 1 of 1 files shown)


MediaRecorder.bs
```diff
@@ -531,6 +543,7 @@ dictionary MediaRecorderOptions {
   unsigned long videoBitsPerSecond;
   unsigned long bitsPerSecond;
   BitrateMode audioBitrateMode = "variable";
+  DOMHighResTimeStamp videoKeyFrameInterval;
```
henbos: Why is this a timestamp and not a double?

Also we need to say that this is in seconds (even if the name doesn't say)

MediaRecorder.bs
```diff
@@ -241,6 +245,14 @@ interface MediaRecorder : EventTarget {
     the value might be surpassed, not achieved, or only be achieved over a long
     period of time.</li>
 
+    <li>Let |videoKeyFrameInterval| be the value of |recorder|'s internal
+    [=[[VideoKeyFrameInterval]]=] slot if defined, and configure the video encoder to
+    target key frame generation with the interval. If the value isn't defined, the
```
henbos: Should we clarify "target"?
Such as, the MediaRecorder SHOULD encode a key frame on the next frame after this interval has passed.
Or should we be vague on purpose?

MediaRecorder.bs
handellm: [`DOMHighResTimestamp`](https://www.w3.org/TR/hr-time/#sec-domhighrestimestamp) is documented to be able to express a duration. Double seems worse because it can be negative or NaN and behavior re that need to be defined. 

Since the type is defined, is there any reason to say it's seconds? It's covered by the `DOMHighResTimestamp` spec.

MediaRecorder.bs
handellm: Good point, updating.

MediaRecorder.bs
henbos: DOMHighResTimestamp is just a typedef for a double. The reason the this type can be used to also describe a time interval is because you can say "interval = timestamp2 - timestamp1" without type conversion.

I have never seen DOMHighResTimestamp used to express an interval when it is an attribute, but getStats() for example is full of uses of double for intervals in seconds.

henbos: Although I see the appeal of ms instead of s, was just thinking of matching with other APIs, but if this is used as intervals in other places then no objection from me

alvestrand:
I think this is OK.
Was wondering whether the interval is reset when the encoder decides on its own to insert a keyframe, but concluded after reading again that the language is sufficiently nonrestrictive that it does not matter.

alvestrand:
Would be good to have 2 slides on this at the February interim, just to make sure it's been on the agenda.

MediaRecorder.bs
alvestrand: performance.now() is probably the most common usage of HiResTimeStamp, and it's an interval (measured from the time origin of the window or worker).
so it's not novel to use it for intervals.

Pehrsons:
To my understanding most encoders configure keyframe intervals in number of frames, not based on time. What's the intended way for an implementation to convert the time interval to a frame interval given that video MediaStreamTracks by definition have no fixed frame rate?

Would it be a better fit to define the keyframe interval as a frame interval? The application might have a better understanding of the rate at which it will feed frames to the MediaRecorder.
Is there specific use case where a time interval is a better fit?

One of the main use cases for recurring keyframes in a non-streamed setting that I can think of is to be able to seek to any given point in the file by jumping to the keyframe prior to that point and decoding your way there. The speed of this process is determined by the number of frames between the keyframe and the point one is seeking to. So to be able to get consistent seek times with a variable frame rate, a key frame interval expressed as a frame interval seems favorable.

A min and a max interval seems like an even better fit for most encoders, but I guess MediaRecorder could be free to infer these.

youennf:
> To my understanding most encoders configure keyframe intervals in number of frames, not based on time.

On macOS/iOS you have the choice.
In many case, there is the possibility to mark a specific frame to be encoded as key frame (see https://www.w3.org/TR/webcodecs/#dictdef-videoencoderencodeoptions for instance).

For what is worth, WebKit is doing the following: a key frame every 2 seconds or 60 frames, whichever comes first.
I wonder whether we could provide suggestions to implementors for the default value, as a non normative statement for instance.

Getting a better sense of the use case to go with either time or/and frame count would be good.

handellm:
Is there an example of an encoder where per-frame control of keyframes isn't possible?

The use case is to let the MediaRecorder support basic control over seekability without requiring post-processing and give the user a basic knob to control the tradeoff between muxed size, quality & seek accuracy. Another useful consequence is also that parts of the stream can be excised with specified temporal accuracy without requiring transcoding.

> Is there specific use case where a time interval is a better fit?

My bet is the user thinks more about seconds than about frame intervals. I struggle to see use cases where a count interval will be better. While it's possible to influence min and max track framerates with `getDisplayMedia`/`getUserMedia` and `applyConstraints`, my understanding is the default source behavior is unspecified. Which means a frame interval gives an undefined temporal accuracy in such a case, or when it's impossible to know the configured rates.

A consequence of a temporal interval is that muxed streams with only keyframes may result. If the output bitrate is constrained, the muxed stream may be of lower quality in stretches with low framerate. But MediaRecorder is supposed to be a simple API. Where more refined control than "increased bitrate config" is needed, the user can always go to WebCodecs.

With the same rationale, I think providing both count & temporal control + logic spec might not be needed for this API.

> I wonder whether we could provide suggestions to implementors for the default value, as a non normative statement for instance.

Do you have a suggestion in mind?

youennf:
> Do you have a suggestion in mind?

I would look at whether Chrome and Firefox have defaults (or will have after this change), similarly to WebKit ones. Maybe we can somehow harmonise values.

handellm:
Yes, that makes sense. 
The current story for Chrome is that there are generally no keyframes forced midstream, except if the mimeType results in getting an accelerated codec, then you get a count based interval of 100 frames (which plays badly with lower framerate input streams). Happy to change this closer to FF/Safari.

MediaRecorder.bs
henbos: Ack

henbos:
I prefer interval over number of frames because I think key frames have a lot to do with seekability, and if this was a count instead of an interval I feel like you'd end up having to calculate it anyway. That said this isn't really my area, so feel free to continue the discussion. Either way I'm happy, +1

henbos (approved):

Pehrsons:
> Is there an example of an encoder where per-frame control of keyframes isn't possible?

Is there a risk with the per-frame API for keyframes, that the encoded stream has more visual artifacts as a result of the encoder having a harder time predicting how to allocate the target bitrate, than if it were to configure the encoder with a per-frame interval?

I have seen behavior in libvpx suggesting this, but I'm not sure how much of it was due to using VPX_EFLAG_FORCE_KF and how much was due to a [libvpx bug](https://bugs.chromium.org/p/webm/issues/detail?id=1714).

IIRC Firefox sets a keyframe interval of 10 seconds, for historical reasons [1], and occasionally recalculates the frame-interval based on recent incoming framerate, so it can configure the encoder with this.

[1] our ebml writer only creates a new cluster on a keyframe, and the max timestamp within a cluster is 32 seconds.

Pehrsons:
> I prefer interval over number of frames because I think key frames have a lot to do with seekability, and if this was a count instead of an interval I feel like you'd end up having to calculate it anyway.

I'm not sure what you mean by seekability, but consider these cases:
- (frame interval) Every chunk contains 1 keyframe and 59 non-keyframes. Seeking to the middle means decoding 30 frames. No matter where in the file you want to seek, the mean number of frames to decode will be 30.
- (time interval) Every chunk is 1 second. One chunk is 30 fps and contains 1 keyframe and 29 non-keyframes. Seeking to the middle means decoding 15 frames. Another chunk is 120fps and contains 1 keyframe and 119 non-keyframes. Seeking to the middle means decoding 60 frames. For any given seek operation, the number of frames to decode will depend on the framerate around that time.

This is the tradeoff I see. Perhaps there's something to be said about predictability and bitrate allocation (and if that is the case, it would be in favor of frame intervals?) but I don't really know.

handellm:
> Is there a risk with the per-frame API for keyframes, that the encoded stream has more visual artifacts as a result of the encoder having a harder time predicting how to allocate the target bitrate, than if it were to configure the encoder with a per-frame interval?

> Is there specific use case where a time interval is a better fit?

It doesn't have anything to do with per-frame API, but with a fixed bitrate budget, quality per frame becomes lower as the ratio of num keyframe/num deltaframes gets higher. When you think about it this is more likely to happen with time-based control with for example low-framerate slideware. With ultimate control over keyframe generation, this seems like a situation a count-based control makes sense.
(Note, ultimate control isn't entirely possible. Encoders anyway emit keyframes when diffs to reference frames becomes too large, so count-based will not be predictable in this situation.)

> This is the tradeoff I see.

Worth mentioning here's it's also the case that with count-based intervals you can't excise with specified accuracy. Say we've recorded a presentation with occasional slide changes. You'd like to save 0.58 minutes - 1.30 minutes in a webm somewhere else. If the framerate was low in the 0.58 section, with count-based intervals chances are you could have to save an area tens of seconds before that point. 

How about specifying both? Say MR configuration is 2 parameters X (min time between keyframes) and Y (min framecount between keyframes):

```
void OnFrame(const VideoFrame& frame) {
  ++currentFrameIndex;
  now = getCurrentTime();
  forceKeyFrame = now - lastKeyFrameTimestamp >= X &&
      currentFrameIndex - lastKeyFrameIndex >= Y;
  if (forceKeyFrame) {
    lastKeyFrameTimestamp = now;
    lastKeyFrameIndex = currentFrameIndex;
  }
  Encode(frame, forceKeyFrame);
}
```

This supports three cases:
1. Pure time-based control (set Y to 0)
2. Pure count-based control (set X to 0)
3. Compromises (X and Y > 0) for variable-rate frame input.

shangl:
Looks good to me. From a user perspective, I'd also say that time-based control would make more sense. Can't speak for what's better with encoders though.

handellm:
Updated to include support for both time and count base interval. PTAL!

MediaRecorder.bs
```diff
@@ -245,15 +249,24 @@ interface MediaRecorder : EventTarget {
...
+    it deems fit.
+    If only |videoKeyFrameIntervalDuration| is defined, the video encoder
+    SHOULD produce a keyframe on the first frame arriving after
+    |videoKeyFrameIntervalDuration| elapsed since the last key frame.
+    If only |videoKeyFrameIntervalCount| is defined, the video encoder
+    SHOULD produce a keyframe on the first frame arriving after
+    |videoKeyFrameIntervalCount| frames passed since the last key frame.
+    If both are defined, the video encoder SHOULD produce a keyframe on the
+    first frame arriving after both |videoKeyFrameIntervalDuration| elapsed AND
+    |videoKeyFrameIntervalCount| frames passed since the last key frame.
```
alvestrand: The text seems to say that if duration is 1 second and count is 10, then a keyframe should be produced after 1 second if the frame rate is > 10, and should be produced after 10 frames if the frame rate is < 10 - "latest one wins".

That seems contrary to the Principle of Least Surprise - should "both" be "either"?

MediaRecorder.bs
```diff
@@ -241,6 +249,25 @@ interface MediaRecorder : EventTarget {
     the value might be surpassed, not achieved, or only be achieved over a long
     period of time.</li>
 
+    <li>Let |videoKeyFrameIntervalDuration| be the value of |recorder|'s internal
+    [=[[VideoKeyFrameIntervalDuration]]=] slot if defined. Additionally let
+    |videoKeyFrameIntervalCount| be the value of |recorder|'s internal
+    [=[[VideoKeyFrameIntervalCount]]=] slot if defined.
+    If neither of the slots are defined, the User Agent may emit key frames as
+    it deems fit.
+    If only |videoKeyFrameIntervalDuration| is defined, the video encoder
```
alvestrand: Add some blank lines to break up this paragraph.

MediaRecorder.bs
handellm: The rationale is to support variable-rate sources where you get low per-frame quality (QP) in stretches where FPS is low with a pure time-based control, since the ratio of keyframes (in which quality is lower) to delta frames (in which quality is enhanced) would be higher in the low FPS stretches. See also earlier comment in https://github.com/w3c/mediacapture-record/pull/216#issuecomment-1415989756.

alvestrand:
Editors meeting: @youennf and @jan-ivar had a few comments (editorial and non-editorial) that need addressing.

youennf:
Two comments:
- It would be good to rephrase the use of the new slots like the other slots in terms of "Constrain the configuration of |recorder| to encode"....
- In macOS and iOS, when setting both values, I think the min is used (I'll check) but the spec is asking for the max.

MediaRecorder.bs
```diff
@@ -241,6 +249,25 @@ interface MediaRecorder : EventTarget {
     the value might be surpassed, not achieved, or only be achieved over a long
     period of time.</li>
 
+    <li>Let |videoKeyFrameIntervalDuration| be the value of |recorder|'s internal
+    [=[[VideoKeyFrameIntervalDuration]]=] slot if defined. Additionally let
+    |videoKeyFrameIntervalCount| be the value of |recorder|'s internal
+    [=[[VideoKeyFrameIntervalCount]]=] slot if defined.
+    If neither of the slots are defined, the User Agent may emit key frames as
+    it deems fit.
```
jan-ivar: Mostly editorial, but this describes runtime behavior as a step in a sequential algorithm, which can be confusing to a reader (e.g. "the User Agent emit key frames" should not be done before the next step here)

MediaCapture recorder already does this in steps above, so this is somewhat of an existing problem, but it would be good to follow the pattern in the step above this one where it uses this form:
> Let audioBitrate be the value of recorder’s [audioBitsPerSecond](https://pr-preview.s3.amazonaws.com/handellm/mediacapture-record/pull/216.html#dom-mediarecorder-audiobitspersecond) attribute, and constrain the configuration of recorder to [do x]

MediaRecorder.bs
handellm: Resolving convo, pls reopen if there are still issues.

MediaRecorder.bs
handellm: Done!

MediaRecorder.bs
handellm: I tried to follow yours and Youenn's comments in 0df3dba, ptal!

handellm:
Youenn, for some reason I can't reply to your comment.
> It would be good to rephrase the use of the new slots like the other slots in terms of "Constrain the configuration of |recorder| to encode"....

Tried to do this in 0df3dba, wdyt?

> In macOS and iOS, when setting both values, I think the min is used (I'll check) but the spec is asking for the max.

The rationale for this spec text is what I discussed with Harald earlier:

"The rationale is to support variable-rate sources where you get low per-frame quality (QP) in stretches where FPS is low with a pure time-based control, since the ratio of keyframes (in which quality is lower) to delta frames (in which quality is enhanced) would be higher in the low FPS stretches. See also earlier comment in https://github.com/w3c/mediacapture-record/pull/216#issuecomment-1415989756"

I realize Apple may not currently be using those rules, but on the other hand the spec includes Apple's implementation by "If neither of the slots are defined, the User Agent may emit key frames as it deems fit.". Wdyt?

handellm:
Friendly ping!

MediaRecorder.bs
```diff
@@ -98,7 +98,15 @@ interface MediaRecorder : EventTarget {
...
+    {{MediaRecorderOptions/bitsPerSecond}} member if it is present, otherwise
+    <code>undefined</code>.</li>
+    <li>Let |recorder| have a <dfn>\[[VideoKeyFrameIntervalDuration]]</dfn> internal
+    slot, initialized to the value of |options|'
+    {{MediaRecorderOptions/videoKeyFrameIntervalDuration}} member if it is present, otherwise
+    <code>undefined</code>.</li>
+    <li>Let |recorder| have a <dfn>\[[VideoKeyFrameIntervalCount]]</dfn> internal
+    slot, initialized to the value of |options|'
+    {{MediaRecorderOptions/videoKeyFrameIntervalCount}} member if it is present, otherwise
     <code>undefined</code>.</li>
```
jan-ivar: While I appreciate the consistency with [[[ConstrainedBitsPerSecond]]](https://w3c.github.io/mediacapture-record/#constrainedbitspersecond) which defaults to `undefined` and then later uses prose like _"If recorder’s [[[ConstrainedBitsPerSecond]]](https://w3c.github.io/mediacapture-record/#constrainedbitspersecond) slot is not `undefined`"_, I think that was a mistake in hindsight.

Implementations tend to be c++, where `undefined` maps poorly. I suggest using `null` instead since you're touching this, if you don't mind.
```suggestion
    <code>null</code>.</li>
    <li>Let |recorder| have a <dfn>\[[VideoKeyFrameIntervalDuration]]</dfn> internal
    slot, initialized to the value of |options|'
    {{MediaRecorderOptions/videoKeyFrameIntervalDuration}} member if it is present, otherwise
    <code>null</code>.</li>
    <li>Let |recorder| have a <dfn>\[[VideoKeyFrameIntervalCount]]</dfn> internal
    slot, initialized to the value of |options|'
    {{MediaRecorderOptions/videoKeyFrameIntervalCount}} member if it is present, otherwise
    <code>null</code>.</li>
```

MediaRecorder.bs
```diff
@@ -241,6 +249,32 @@ interface MediaRecorder : EventTarget {
     the value might be surpassed, not achieved, or only be achieved over a long
     period of time.</li>
 
+    <li>Let |videoKeyFrameIntervalDuration| be the value of |recorder|'s internal
+    [=[[VideoKeyFrameIntervalDuration]]=] slot if defined, and let
+    |videoKeyFrameIntervalCount| be the value of |recorder|'s internal
+    [=[[VideoKeyFrameIntervalCount]]=] slot if defined. The UA SHOULD constrain the
```
jan-ivar: I couldn't find any precedent for using _"if defined"_ to mean _"not `undefined`"_, so I'd like to avoid introducing that pattern here. Switching to `null` default values above should solve this. 

This should also simplify these assignments, which otherwise appear to be partially set:
```suggestion
    <li>Let |videoKeyFrameIntervalDuration| be
    |recorder|.[=[[VideoKeyFrameIntervalDuration]]=], and let
    |videoKeyFrameIntervalCount| be
    |recorder|.[=[[VideoKeyFrameIntervalCount]]=]. The UA SHOULD constrain the
```

MediaRecorder.bs
```diff
@@ -241,6 +249,32 @@ interface MediaRecorder : EventTarget {
...
+    configuration of |recorder| so that the video encoders follow the below rules:
+
+    <ol>
```
jan-ivar: I think unnumbered steps would be better here since they aren't sequential.
```suggestion
    <ul>
```

MediaRecorder.bs
```diff
@@ -241,6 +249,32 @@ interface MediaRecorder : EventTarget {
...
+    <li>If neither of the slots are defined, the User Agent may emit key frames as
+    it deems fit.</li>
```
jan-ivar: For consistency I think we should reference the let variables not "the slots". if it's unclear what "neither" refers to, I wonder if it might work to move it to the bottom (after "both")?

jan-ivar (commented):
Thanks for the updates! Sorry I noticed more nits I didn't catch the first time, but overall I think this looks good!

MediaRecorder.bs
handellm: Done.

MediaRecorder.bs
handellm: Done.

handellm:
Thanks for comments! New iteration uploaded, PTAL again.

MediaRecorder.bs
```diff
@@ -220,8 +228,8 @@ interface MediaRecorder : EventTarget {
     media type, container, and codec specified in the [=[[ConstrainedMimeType]]=]
     slot.</li>
 
-    <li>If |recorder|'s [=[[ConstrainedBitsPerSecond]]=] slot is not
-    <code>undefined</code>, set |recorder|'s {{MediaRecorder/videoBitsPerSecond}} and
+    <li>If |recorder|'s [=[[ConstrainedBitsPerSecond]]=] slot is
```
jan-ivar: Looks like an accidental flip
```suggestion
    <li>If |recorder|'s [=[[ConstrainedBitsPerSecond]]=] slot is not
```

MediaRecorder.bs
```diff
@@ -241,6 +249,33 @@ interface MediaRecorder : EventTarget {
...
+    seconds elapsed since the last key frame.</li>
+
+    <li>If |videoKeyFrameIntervalCount| is not <code>null</code> and |videoKeyFrameIntervalDuration| is <code>null</code>,
+    the video encoder produces a keyframe on the first frame arriving after |videoKeyFrameIntervalCount|
+    frames passed since the last key frame.</li>
+
+    <li>If |videoKeyFrameIntervalDuration| and |videoKeyFrameIntervalCount| are both not <code>null</code>,
+    the video encoder produces a keyframe on the first frame arriving after
+    both |videoKeyFrameIntervalDuration| seconds elapsed AND at least |videoKeyFrameIntervalCount|
+    frames passed since the last key frame.</li>
```
jan-ivar: Thanks, this seems clear!

I guess my remaining question is a content one, for folks like @Pehrsons: is this "both" behavior desirable and usable? The alternative would be to throw if both are provided (one less thing to test). Maybe this was already covered in discussions and I missed it.

In any case I'm not gonna block on it.

jan-ivar (approved):
LGTM modulo accidental flipping of [[ConstrainedBitsPerSecond]].

henbos (approved):

MediaRecorder.bs
handellm: Thanks! All done now I hope. 

The rationale was discussed with Harald before on this PR:

"The rationale is to support variable-rate sources where you get low per-frame quality (QP) in stretches where FPS is low with a pure time-based control, since the ratio of keyframes (in which quality is lower) to delta frames (in which quality is enhanced) would be higher in the low FPS stretches. See also earlier comment in https://github.com/w3c/mediacapture-record/pull/216#issuecomment-1415989756"

handellm:
Youenn, any thoughts from you?

MediaRecorder.bs
Pehrsons: I don't have much thought on using both. I guess mainly because I don't really see a use case for a time-based keyframe interval. I've only seen mentioned the use case in [this comment](https://github.com/w3c/mediacapture-record/pull/216#issuecomment-1415989756) and think that with a framecount-based interval one could achieve similar results by having identical (or [negative](https://www.ietf.org/archive/id/draft-ietf-cellar-matroska-16.html#section-11.2-8)) presentation timestamps for the necessary frames before the desired cut without too much storage overhead, since the overhead would only consist of non-keyframes.

I guess a time-based interval could still make sense if there is precedence in the industry, which there seems to be.
As far as a use-case for setting both framecount- and time-based intervals goes, not sure.

@youennf [might](https://github.com/w3c/mediacapture-record/pull/216#issuecomment-1441955962) have more thoughts.

youennf: I would tend to restrict to either one of the two options for now (throw if both are set or ignore one if the other is there).
This restriction does not seem to impact the use cases we care.
It will be easy to remove this restriction later on if we see value in this.

handellm: @Pehrsons 

> IIRC Firefox sets a keyframe interval of 10 seconds, for historical reasons [1], and occasionally recalculates the frame-interval based on recent incoming framerate, so it can configure the encoder with this.

> I guess mainly because I don't really see a use case for a time-based keyframe interval.

These statements seem a bit contradictory (time constraints imposed by the muxer implementation are mentioned. When the user has such time constraints to avoid transcoding the stream, a sole frame based control would force the need for framerate estimation or assumptions in JS (is this always possible?).

The "both" case was introduced in this discussion as a simple quality compromise solution for dealing with variable framerate. Since I have no live request for this feature right now I'm happy to punt that for later. I uploaded a new version which ignore either if both are set.

Pehrsons: In Firefox I intend to (eventually) allow the muxer to write a new cluster without a keyframe, thereby removing the need for the encoder to use a time-based key frame interval. I hope this resolves the contradiction.

MediaRecorder.bs
```diff
@@ -241,6 +249,30 @@ interface MediaRecorder : EventTarget {
...
 
+    <li>Let |videoKeyFrameIntervalDuration| be
+    |recorder|.[=[[VideoKeyFrameIntervalDuration]]=], and let
+    |videoKeyFrameIntervalCount| be
+    |recorder|.[=[[VideoKeyFrameIntervalCount]]=]. The UA SHOULD constrain the
+    configuration of |recorder| so that the video encoder follows the below rules:
+    <ul>
+    <li>If |videoKeyFrameIntervalDuration| is not <code>null</code> and |videoKeyFrameIntervalCount| is <code>null</code>,
+    the video encoder produces a keyframe on the first frame arriving after |videoKeyFrameIntervalDuration|
+    seconds elapsed since the last key frame.</li>
```
youennf:
```suggestion
    milliseconds elapsed since the last key frame.</li>
```

youennf (approved):

henbos (approved):

jan-ivar (approved):

jan-ivar:
Thanks for catching the seconds! Approved with milliseconds to follow [§ 8.3. Use milliseconds for time measurement](https://w3ctag.github.io/design-principles/#milliseconds) and for consistency with existing timeslice.

PR #223, opened 2024-09-05
Update https://w3c.github.io/mediacapture-record/#dom-blobeventinit-timecode.

### LINKED ISSUE #222: MediaRecorder BlobEvent timecode clarification
From:  https://w3c.github.io/mediacapture-record/#dom-blobeventinit-timecode

> `timecode`, of type `DOMHighResTimeStamp`, readonly
> The difference between the timestamp of the first chunk in data and the timestamp of the first chunk in the first BlobEvent produced by this recorder as a DOMHighResTimeStamp [HR-TIME]. Note that the timecode in the first produced BlobEvent does not need to be zero.

I don't think I've ever fully understand this sentence.  If this is the value of the timestamp of the first chunk in `data`, compared to the value of the first chunk in `data` of the first event produced by this recorder, and if this is the first event, how could the value of `timecode` not be zero `0`?

I believe at some point in the past, `timecode` would actually start at zero `0` for the first event, and then subsequent `dataavailable` `BlobEvent` `timecode` could be used to determine the duration of recorded MediaStream data thus far.  It seems that the current way is to note the first event's `timecode` and offset subsequent `timecode` events by that first event's `timecode` value to see how much time has passed.  That's fine, but depends on the assumption that the first event's `timecode` is the zero relative to other events.

handellm:
The spec text was unclear. From offline discussion with the original spec author it's believed that the initial intent of the spec language was meant to convey that the first produced timecode would contain an absolute timestamp. However, it's not clear what use case is unlocked by that, as the user can construct their own absolute timestamp with performance.now() on event reception. This ignores the fact that the event handler can execute delayed, but since it's not clear what problem this fixes, this PR simply makes the timestamp sequence start with zero instead.

Fixes #222.

***
<a href="https://pr-preview.s3.amazonaws.com/handellm/mediacapture-record/pull/223.html" title="Last updated on Sep 26, 2024, 2:03 PM UTC (5cf4f29)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/mediacapture-record/223/48e2abf...handellm:5cf4f29.html" title="Last updated on Sep 26, 2024, 2:03 PM UTC (5cf4f29)">Diff</a>

### DIFF
--- a/MediaRecorder.bs
+++ b/MediaRecorder.bs
@@ -680,10 +680,11 @@ enum RecordingState {
   encoding of the blob data.
 
   <dt><dfn attribute for="BlobEvent"><code>timecode</code></dfn></dt>
-  <dd>The difference between the timestamp of the first chunk in
-  {{BlobEvent/data}} and the timestamp of the first chunk in the first
-  {{BlobEvent}} produced by this recorder as a {{DOMHighResTimeStamp}} [[!HR-TIME]].  Note that the {{BlobEvent/timecode}}
-  in the first produced {{BlobEvent}} does not need to be zero.
+  <dd>For a MediaRecorder instance, the {{BlobEvent/timecode}} in the first
+  produced {{BlobEvent}} contains 0. Subsequent {{BlobEvent}}'s
+  {{BlobEvent/timecode}} contain the difference of the timestamp of creation of
+  the first chunk in said {{BlobEvent}} and the timestamp of the first chunk of the
+  first produced {{BlobEvent}}, as {{DOMHighResTimeStamp}} [[!HR-TIME]].
   </dd>
 
 </dl>
--- a/MediaRecorder.bs
+++ b/MediaRecorder.bs
@@ -681,7 +681,7 @@ enum RecordingState {
 
   <dt><dfn attribute for="BlobEvent"><code>timecode</code></dfn></dt>
   <dd>For a MediaRecorder instance, the {{BlobEvent/timecode}} in the first
-  produced {{BlobEvent}} contains 0. Subsequent {{BlobEvent}}'s
+  produced {{BlobEvent}} MUST contain 0. Subsequent {{BlobEvent}}'s
   {{BlobEvent/timecode}} contain the difference of the timestamp of creation of
   the first chunk in said {{BlobEvent}} and the timestamp of the first chunk of the
   first produced {{BlobEvent}}, as {{DOMHighResTimeStamp}} [[!HR-TIME]].


bradisbell:
> From offline discussion with the original spec author it's believed that the initial intent of the spec language was meant to convey that the first produced timecode would contain an absolute timestamp.

@handellm Actually, that makes sense... that way, the user can know when recording started.

We know from MediaRecorder's `start` event *roughly* what browser time the recorder started, but we *don't* know the browser timestamp of the first recorded audio/video data.

So, if a user clicks record and we start recording a MediaStream with a MediaRecorder, with a `timeslice` of `5_000`...

 - `2024-09-05 15:19:32.500Z` We ask MediaRecorder to start
 - `2024-09-05 15:19:33.000Z` MediaRecorder `start` fires
 - `2024-09-05 15:19:33.100Z` The next video frame comes in... this is the first frame since the MediaRecorder has started, and suppose it's a screen recording so frames are random and sparse.
 - ... more frames here ...
 - `2024-09-05 15:19:38.222Z` MediaRecorder `dataavailable` fires.  This is the first chunk.

Now, in this sequence we know when we asked MediaRecorder to start, but this isn't a good timestamp to use for the actual start time because it can take quite a bit of time for MediaRecorder to get moving, especially with hardware codecs involved that need to be woken up and initialized.  MediaRecorder tells us when it starts, but that isn't a good timestamp either because it isn't relative to the MediaStream.  There could be several seconds of nothing before the next frame.  Only when `dataavailable` fires do we know what was captured and when.

I might just be hopeful here, but if the original intent in the spec was to give a timestamp for that first recorded data... maybe we should find a way to make that clearer rather than abandon it?  Or, if this is an expansion of the API and not how browsers work today, then maybe this is worthy of a future addition?

MediaRecorder.bs
```diff
@@ -680,10 +680,11 @@ enum RecordingState {
   encoding of the blob data.
 
   <dt><dfn attribute for="BlobEvent"><code>timecode</code></dfn></dt>
-  <dd>The difference between the timestamp of the first chunk in
-  {{BlobEvent/data}} and the timestamp of the first chunk in the first
-  {{BlobEvent}} produced by this recorder as a {{DOMHighResTimeStamp}} [[!HR-TIME]].  Note that the {{BlobEvent/timecode}}
-  in the first produced {{BlobEvent}} does not need to be zero.
+  <dd>For a MediaRecorder instance, the {{BlobEvent/timecode}} in the first
+  produced {{BlobEvent}} contains 0. Subsequent {{BlobEvent}}'s
```
jan-ivar: @Pehrsons does this match what Firefox does?

Pehrsons: We don't implement `timecode`: https://searchfox.org/mozilla-central/rev/a13db27562f9237db97e2ea5b01dc879d5b55b74/dom/webidl/BlobEvent.webidl

alvestrand:
```suggestion
  produced {{BlobEvent}} MUST contain 0. Subsequent {{BlobEvent}}'s
```

jan-ivar:
```suggestion
  produced {{BlobEvent}} MUST be 0. Subsequent {{BlobEvent}}'s
```

jan-ivar (approved):
+1 with nit

handellm:
@bradisbell - if the time of arrival of the first frame is to be timestamped, there are better ways of observing this such as using MediaStreamTrackProcessor/generator. I'm going to merge this PR now and follow up with WPTs and Chromium change.

alvestrand:
Added "editors can integrate" label to PR too (it was wrongly applied to just the issue).

